PR #9, opened 2015-12-27
Fix java compatibility to include into Android projects

thuytrinh:
Same as https://github.com/rharter/auto-value-moshi/pull/5


rharter:
Sorry I missed this, @thuytrinh, but it's already been fixed in master.

thuytrinh:
Great!

PR #10, opened 2015-12-27
Remove param Foo.class from registerTypeAdapterFactory

thuytrinh:
`registerTypeAdapterFactory` only takes one param according to [Gson Javadocs](http://google.github.io/gson/apidocs/).


rharter:
Good catch, thanks.

PR #12, opened 2016-03-07
Use primitive types

solcott:
Use primitive types when possible.  This fixes NullPointerExceptions when properties that are primitive types are not in the supplied json

### DIFF
--- a/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -2,9 +2,12 @@ package com.ryanharter.auto.value.gson;
 
 import com.google.auto.service.AutoService;
 import com.google.auto.value.extension.AutoValueExtension;
+import com.google.common.base.Defaults;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
+import com.google.common.primitives.Primitives;
 import com.google.gson.Gson;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -347,6 +359,34 @@ public MethodSpec createReadMethod(ClassName className,
...
+                valueString = valueString + "f";
+                break;
+              case "long":
+                valueString = valueString + "L";
+            }
+          }
+        }
+      }
+
+    } catch (Exception e) {
```
JakeWharton: Why is this here with such a wide exception type?

src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -347,6 +359,34 @@ public MethodSpec createReadMethod(ClassName className,
...
+          if (!Strings.isNullOrEmpty(valueString)) {
+            switch (type.toString()) {
+              case "double":
+                valueString = valueString + "d";
+                break;
+              case "float":
+                valueString = valueString + "f";
+                break;
+              case "long":
+                valueString = valueString + "L";
```
JakeWharton: All `case` branches should include a `break;`

solcott: Fixed

src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
solcott: Changed this to just catch ClassNotFoundException

src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -296,8 +299,17 @@ public MethodSpec createReadMethod(ClassName className,
     for (Property prop : adapters.keySet()) {
       FieldSpec field = FieldSpec.builder(prop.type, prop.name).build();
       fields.put(prop, field);
-
-      readMethod.addStatement("$T $N = null", field.type.isPrimitive() ? field.type.box() : field.type, field);
+      if(field.type.isPrimitive()){
```
rharter: Please fix up the code format.  There seem to be inconsistent whitespace around the if/else blocks.

solcott: Fixed

rharter:
Looks good. :+1:

Issue #17, opened 2016-03-24
Support mechanism to exclude some classes


lenguyenthanh:
Default behavior of this extension is create TypeAdapterFactory for every AutoValue classes. But maybe not all of AutoValue classes need it. So it'd be great if we have a mechanism to exclude or include a specific AutoValue class. A @Gson Annotation is a good idea.

lenguyenthanh:
@rharter do you have any idea for this feature? is it good? I'm happy to make a PR. But wanna discuss with you first. Cheers!

gabrielittner:
I've just found another reason for this. The gson and moshi extensions don't support generics in AutoValue classes and generate code that won't compile for those. So without opt-in/out you either can't use generics or can't use the gson/moshi extension.

rharter:
I do think this is a good idea, I'm just trying to come up with a good way to handle it.  We need to play nice with other extensions, so I'd love it if there was a way to include/exclude classes without the need for another extension, but I'm not sure there is.

Are you suggesting an explicit `@Gson` annotation for inclusion?

```java
@AutoValue @Gson public abstract class Foo {

}
```

lenguyenthanh:
Yes, it is exactly what I propose. I'd like to add a runtime Module for `@Gson` annotation only. On main extension library, it should start generate code only if a class has `@Gson` annotation.

gabrielittner:
I think there are two completely different usage scenarios. You either want it to run on all and exclude few or the opposite. Adding `@Gson` to all your classes can get very annoying and using the static `TypeAdapterFactory` method as opt-in signal won't work with auto-registration.
 What do you think of having two separate artifacts? One works like the current one and has a `@NoGson` annotation to opt-out. The other one requires explicit opt-in by adding the static `TypeAdapterFactory` method or adding `@Gson`.

rharter:
First of all, there's no need for a runtime module, as even if there was a `@Gson` annotation, it would only need source retention and not be needed at runtime.

That being said, @gabrielittner's points are exactly my concerns.  In your case, you want to exclude most classed, but explicitly include certain ones.  Most of my cases are the opposite, in which most of my classes, benefit from having TypeAdapters. The challenge is that making things easier for one case makes them harder for the other.

As for the idea of having multiple artifacts, that sounds like a management headache and would be confusing to users, so I'm not a big fan of that idea.

lenguyenthanh:
[...]

PR #22, opened 2016-04-07
Adds requirement of public static method returning TypeAdapter<Foo> to apply the extension.

rharter:
This allows you to only generate TypeAdapters for applicable classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,31 +6,37 @@ An extension for Google's [AutoValue](https://github.com/google/auto) that creat
 
 ## Usage
 
-Simply include auto-value-gson in your project and add the generated Serializer and Deserializer as a TypeAdapter.  You can also annotate your properties using `@SerializedName` to define an alternate name for de/serialization.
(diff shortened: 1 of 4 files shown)


JakeWharton:
Is there a test asserting the type adapter isn't generated unless there's the method?

Also do you think that if there's a `TypeAdapter`-returning `public static` method with the wrong generic type that a warning should be emitted? This could happen if you copy/paste a class and change the name & properties but forget to change the method.

README.md
```diff
@@ -6,32 +6,38 @@ An extension for Google's [AutoValue](https://github.com/google/auto) that creat
...
+}
+
+public class AutoValueTypeAdapterFactory extends TypeAdapterFactory {
+  public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+    Class<? super T> rawType = type.getRawType();
+    if (rawType.equals(Foo.class)) {
+      return new Foo.typeAdapter(gson);
+    } else if (rawType.equals(Bar.class)) {
+      return new Bar.typeAdapter(gson);
+    }
```
JakeWharton: Missing `return null;`

rharter:
This closes #17

PR #24, opened 2016-04-12
Readme update

xMikeTx:
Update Readme. Because it currently does not work with version 0.2.2 of the processor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Simply include auto-value-gson in your project and add a public static method to
   // The public static method returning a TypeAdapter<Foo> is what
   // tells auto-value-gson to create a TypeAdapter for Foo.
   public static TypeAdapter<Foo> typeAdapter(Gson gson) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,13 @@ apt 'com.ryanharter.auto.value:auto-value-gson:0.2.2'
 
 (Using the [android-apt](https://bitbucket.org/hvisser/android-apt) plugin)
 
+In Order for the latest update to work (where this documentation is referring to) you need to have the snapshot
```
rharter: Please remove the Jitpack reference.  There will always be certain features that are only available in snapshot builds, but we can't start adding (and updating) jitpack commit specific dependencies.

vdubedout:
Finally got it working with your readme, Thanks Mike :)

README.md
xMikeTx: Make sense, but I the documentation should refer to a valid gradle dependency, which works with the provided example.

PR #32, opened 2016-04-20
fix example in README

ndurell:
This tripped me up when I was trying to use this library.
In the latest gson it looks like TypeAdapterFactory is an interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Simply include auto-value-gson in your project and add a public static method to
 }
 
 public class AutoValueTypeAdapterFactory implements TypeAdapterFactory {
+  @Override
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ Simply include auto-value-gson in your project and add a public static method to
   }
 }
 
-public class AutoValueTypeAdapterFactory extends TypeAdapterFactory {
+public class AutoValueTypeAdapterFactory implements TypeAdapterFactory {
   public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
```
JakeWharton: Add `@Override` annotation while you're here?

ndurell:
@JakeWharton Added Override.

rharter:
Thanks! 👍

PR #33, opened 2016-04-25
Class is final and does not override equals

PaulWoitaschek:
So == can be used.


rharter:
Thanks for the contribution.  This was only ever temporary in the readme, and will be removed with PR #37, so I'm going to close this.

PR #36, opened 2016-05-08
Fix generation of TypeAdapter.read() for 'char' values

dimsuz:
This pull request fixes a TypeAdapter generation for cases when annotated class has a char field.
At the moment incuding `char` field results in generation of TypeAdapter that fails to compile.

To reproduce:
```java
@AutoValue
public abstract class Test {
  public abstract char c();
  public static TypeAdapter<Test> typeAdapter(Gson gson) {
    return new AutoValue_Test.GsonTypeAdapter(gson);
  }
}
```

Above code will result in generation of `TypeAdapter.read()` which fails to compile.
What happens is that `auto-value-gson` will use Guava's `Defaults.getDefault(char.class)`, which returns `\0`, then apply `.toString()` to it and write it **unquoted** as `char c = \0`, which will break the compilation.

This PR wraps `\0` in single quotes which fixes the issue.

### DIFF
--- a/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -383,6 +383,9 @@ public class AutoValueGsonExtension extends AutoValueExtension {
               case "long":
                 valueString = valueString + "L";
                 break;
+              case "char":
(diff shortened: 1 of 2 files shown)


dimsuz:
By the way, I didn't check yet whether Moshi extension has the same issue. This one caught me after I started using `auto-value-gson` in one of the large projects I work on (during transition from `@AutoParcel` to `@AutoValue`)

rharter:
This looks great, thanks!

PR #37, opened 2016-05-11
Adds type adapter factory for all generated type adapters.

### LINKED ISSUE #19: Support single TypeAdapterFactory for all AutoValue generated type adapters.
The currently TypeAdapterFactory mechanism requires users to add each individual TypeAdapterFactory to each Gson instance in order to de/serialize objects.  This can be cumbersome, especially when you have a large number of model objects.  

As long as we're already generating code, we could alleviate this by generating a single `AutoValueTypeAdapterFactory` that will serve as TypeAdapterFactory for all AutoValue: Gson generated type adapters.  This can use the standard TypeAdapterFactory mechanism to check the class type and delegate the request to the appropriate TypeAdapter.

The challenge I see here is that the AutoValue Extension mechanism processes classes serially, meaning this would require an additional annotation process that processes the same `@AutoValue` annotation, checks applicability, and generates a single `AutoValueTypeAdapterFactory` for all applicable classes.

This would reduce the code needed to configure your Gson instance to the following:

```java
Gson gson = new GsonBuilder()
  .registerTypeAdapterFactory(new AutoValueTypeAdapterFactory())
  .build();
```

rharter:
Fixes #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ An extension for Google's [AutoValue](https://github.com/google/auto) that creat
 
 ## Usage
 
-Simply include auto-value-gson in your project and add a public static method to your `@AutoValue` annotated class returning a TypeAdapter.  You can also annotate your properties using `@SerializedName` to define an alternate name for de/serialization.
+Simply include auto-value-gson in your project and add a public static method to your `@AutoValue` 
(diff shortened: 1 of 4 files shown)


hzsweers:
Lgtm

PR #42, opened 2016-05-17
Replaced equals check with isAssignableFrom in  AutoValueGsonTypeAdapterFactory

fabioCollini:
Hi, I am using this framework and it's great!
I have found a small problem in AutoValueGsonTypeAdapterFactory. Sometimes, for example in nested objects, the second parameter of `AutoValueGsonTypeAdapterFactory.create` contains the generated class and not the original one. I have changed the check from `rawType.equals(Person.class)` to `Person.class.isAssignableFrom(rawType)` to fix it.
I updated the example to add a nested object `Address` with a `SerializedName` annotated field, with the equals check it doesn't work (the autogenerated `TypeAdapter<Address>` is not used).


rharter:
Looks great.  Thanks!

vanniktech:
@rharter this should also be changed in the auto value moshi project, right?

rharter:
@vanniktech It's already done :)

PR #56, opened 2016-06-16
Suppress unchecked warnings on generated class

igoticecream:
Hello.

I made a little addition to this library. Javac gives me unchecked warnings when compiling using auto-value-gson.

I'd like to keep my build clean (0 warnings) as possible.

Thanks.


rharter:
I think this looks good, you just need to update the tests to expect the `@SuppressWarnings` annotation and things should be great, if you want to reopen and finish it off.

PR #59, opened 2016-07-21
Switch to autovalue-esque style of generating Factories

### LINKED ISSUE #52: ZipException in IntelliJ/Android Studio multiple modules
If I have two IntelliJ/Studio modules, both containing AutoValue classes, and one depends on the other. The generated `com.ryanharter.auto.value.gson.AutoValueGsonTypeAdapterFactory` class is present in both the modules and this causes a ZipException when building the project.

### LINKED ISSUE #58: AutoValueGsonAdapterFactoryProcessor doesn't account for projects with multiple modules
AutoValueGsonAdapterFactoryProcessor generated a class called com.ryanharter.auto.value.gson.AutoValueGsonTypeAdapterFactory.  This obviously doesn't work when auto-value-gson is used in multiple modules.  It would be nice if auto-value-gson would either allow modules to change the package/class name via an annotation processor argument or have some other way to workaround this.

hzsweers:
Followup from https://github.com/rharter/auto-value-moshi/pull/47 and resolves #52 and resolves #58. 

(copying over from https://github.com/rharter/auto-value-moshi/pull/47)

This would be a breaking change, but for the better I think as it'd be a bit muddled to try to maintain the old behavior and this one's as well.

Now, people would have to declare their own @GsonTypeAdapterFactory class(es) to specify where they want it generated.

```java
@GsonTypeAdapterFactory
public abstract class MyAdapterFactory implements TypeAdapterFactory {
  // Static factory method to access the package
  // private generated implementation
  public static TypeAdapterFactory create() {
    return new AutoValueGson_MyAdapterFactory();
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,37 @@ final Gson gson = new GsonBuilder()
 
 Now build your project and de/serialize your Foo.
 
-In addition to generating implementations of your `@AutoValue` annotated classes, auto-value-gson
-also generates an `AutoValueGsonTypeAdapterFactory` class which you can register with your 
-`GsonBuilder` to automatically add all of your generated TypeAdapters.
+## Factory
+
+Optionally, auto-value-gson can create a single [TypeAdapterFactory](https://google.github.io/gson/apidocs/com/google/gson/TypeAdapterFactory.html) so
+that you don't have to add each generated TypeAdapter to your Gson instance manually.
+
+To generate a `TypeAdapterFactory` for all of your auto-value-gson classes, simply create
+an abstract class that implements `TypeAdapterFactory` and annotate it with `@GsonTypeAdapterFactory`,
+and auto-value-gson will create an implementation for you.  You simply need to provide a static
+factory method, just like your AutoValue classes, and you can use the generated `TypeAdapterFactory`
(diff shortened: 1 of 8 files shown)


JakeWharton:
> This would be a breaking change

Change `gradle.properties` to 0.4.0-SNAPSHOT?

hzsweers:
I usually leave that to the library maintainer, but I can bump it here if @rharter's ok with it

I missed that there was a sample project, but can't get it to recognize the `@GsonTypeAdapterFactory` annotation on the compile classpath. Are we sure that keeping it in the same module as the processor will work?

rharter:
You can bump the version now to a new snapshot.

hzsweers:
any thoughts on the annotation issue? Wondering if it's maybe just an issue with the local sample, unless this has been reported in the equivalent change in auto-value-moshi

rharter:
So I think you would need the `auto-value-gson` dependency added to `compileOnly` in addition to the `apt` configuration.  That should make it visible in the source.  For Android we'd need this dependency added to the `provided` configuration, which should probably be in the README.

PR #61, opened 2016-07-29
Adds support for generic types. fixes #35

### LINKED ISSUE #35: Generics are not supported
I am keeping this issue short as I notice in https://github.com/rharter/auto-value-gson/issues/17#issuecomment-203367468 that there is awareness about it.

The main reason I open this issue is to confirm the same.

rharter:


example/src/test/java/com/ryanharter/auto/value/gson/example/WebResponseTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+
+      User user = (User) o;
+
+      if (firstname != null ? !firstname.equals(user.firstname) : user.firstname != null)
+        return false;
+      return lastname != null ? lastname.equals(user.lastname) : user.lastname == null;
```
JakeWharton: If only there was a library to write this for you...

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -2,7 +2,9 @@
 
 import com.google.auto.value.processor.AutoValueProcessor;
 import com.google.common.collect.ImmutableSet;
+import com.google.gson.reflect.TypeToken;
 import com.google.testing.compile.JavaFileObjects;
+import java.lang.reflect.ParameterizedType;
```
JakeWharton: nit: these are unused

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -548,4 +547,125 @@ public void simple() {
...
+        + "import java.lang.String;\n"
+        + "import java.lang.reflect.ParameterizedType;\n"
+        + "import java.lang.reflect.Type;\n"
+        + "\n"
+        + "final class AutoValue_Foo<A, B, C> extends $AutoValue_Foo<A, B, C> {\n"
+        + "  AutoValue_Foo(C c, A a, B b, String d) {\n"
+        + "    super(c, a, b, d);\n"
+        + "  }\n"
+        + "\n"
+        + "  public static final class GsonTypeAdapter extends TypeAdapter<Foo> {\n"
```
JakeWharton: I think @swankjesse mentioned this in Slack, but this really seems like it wants to be
```java
public static final class GsonTypeAdapter<A, B, C> extends TypeAdapter<Foo<A, B, C>>
```
and having a constructor of
```java
GsonTypeAdapter(Gson gson, TypeAdapter<? extends A> aAdapter, TypeAdapter<? extends B> bAdapter, TypeAdapter<? extends C> cAdapter)
```
This requires also that there's a `TypeAdapterFactory` which can do the ugly reflection on the runtime type for the actual generics.

However, currently you're already forcing the user to create a `TypeToken` for (in this case) `Foo` with all of it's generics filled out for registering with the Gson instance. This means that they're effectively fixed at compile-time and not truly generic like a `TypeAdapterFactory` would be.

(It's a shame I didn't look at this when we were actually in the same room instead of now having to talk about this 1000 miles apart)

JakeWharton: Actually it looks like there is a `TypeAdapterFactory` generated, it's just omitted from the tests.

I think this could be fixed then by changing
```java
public static TypeAdapter<Foo> typeAdapter(Gson gson, TypeToken<? extends Foo> typeToken)
```
to
```java
public static <A, B, C> TypeAdapter<Foo<A, B, C>> typeAdapter(Gson gson, TypeToken<? extends Foo<A, B, C>> typeToken)
```
and then properly hoisting the generics onto `GsonTypeAdapter`.

example/src/test/java/com/ryanharter/auto/value/gson/example/WebResponseTest.java
rharter: I was thinking that would invalidate the test, since Gson would have to use the generated `TypeAdapter` for this class, but that shouldn't effect things since I'm still comparing to known values.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -548,4 +545,117 @@ public void simple() {
...
+        + "            d = dAdapter.read(jsonReader);\n"
+        + "            break;\n"
+        + "          }\n"
+        + "          default: {\n"
+        + "            jsonReader.skipValue();\n"
+        + "          }\n"
+        + "        }\n"
+        + "      }\n"
+        + "      jsonReader.endObject();\n"
+        + "      return new AutoValue_Foo(c, a, b, d);\n"
```
JakeWharton: This should have `<>` since it's a generic type in this case.

JakeWharton:
Awesome

PR #66, opened 2016-08-11
Handle parameterized generic fields

### LINKED ISSUE #65: Generics with parameterized fields
A model with a parameterized field such as this one:

``` java
@AutoValue
public abstract class ListResponse<T> {
  public abstract List<T> results();
}
```

Will produce:

``` java
this.resultsAdapter = gson.getAdapter(new TypeToken<List<T>>(){});
```

This doesn't work since `T` is erased during runtime.

oguzbabaoglu:
Fixes #65 and #68


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -310,16 +312,51 @@ public TypeSpec createTypeAdapter(ClassName className, ClassName autoValueClassN
...
   }
 
+  private static void buildParameterizedToken(MethodSpec.Builder constructor, TypeName typeArg,
+                                              List<TypeVariableName> typeParams, boolean first) {
+    if (!first) {
+      constructor.addCode(", ");
+    }
+    if (typeArg instanceof ParameterizedTypeName) {
+      ParameterizedTypeName paramTypeArg = (ParameterizedTypeName) typeArg;
+      constructor.addCode("$T.newParameterizedTypeWithOwner(null, $T.class", $Gson$Types.class, paramTypeArg.rawType);
```
oguzbabaoglu: I am not really happy with using Gson's internal api here. Anyone have a better idea?

evant: Copy it out for yourself, that's what I did https://github.com/evant/gsonvalue/blob/master/gsonvalue/src/main/java/me/tatarka/gsonvalue/internal/Types.java

oguzbabaoglu: Yup that is a straightforward solution but this library doesn't have a runtime component to put that in. We would either need to add a runtime module or generate this helper class in compile time.

rharter: @swankjesse Do you mind providing some input here?  If someone is using Gson can we consider it safe to use the internal `$Gson$Types` class, since we depend on a specific version of Gson?

JakeWharton: You depend on a version but your users might be using a newer (or even
older) version. That class could be removed in the next patch release. It
won't, but it could.

rharter: Yeah, my question was basically "is this protected for patch releases", but I suppose since it's not public API it's not.

JakeWharton: Yeah. Maybe we could add the factory methods to TypeToken though. Moshi
exposes Types for this which is very convenient.

JakeWharton: https://github.com/google/gson/pull/900

oguzbabaoglu:
This now depends on having gson 2.8.0 as a runtime dependency in the users project for handling parameterized types. For already existing cases using an older gson dep should not break anything.

rharter:
Awesome, this is great, thanks! I think the dependency on 2.8.0 specifically warrants something in the readme (I'm anticipating issues filed for this :) ).  Would you mind adding something in?

oguzbabaoglu:
np! I added some info to README under a new title "Generics support". should be good to go.

kibao:
@rharter Could you release snapshot with this feature or a new version of library ?

PR #71, opened 2016-10-13
Initialize known collection types to empty instead of null

cushon:


JakeWharton:
What about `@Nullable` properties. Should we allow them to be initialized to `null` since you've essentially opted into the behavior?

cushon:
> Should we allow them to be initialized to null since you've essentially opted into the behavior?

SGTM, I updated the pull request.

rharter:
Thanks @cushon, this looks great.  I was initially thrown by the `Immutable*` additions, but that will only become a dependency if the app is already using Guava, so nothing is imposed on the user.

rharter:
@hzsweers has raised a valid concern that this breaks the AutoValue expectation that null collection and empty collections are different.  If your collection property is non-nullable, then it should exist in your JSON.  (i.e. if you deserialize JSON with this change, then serialize JSON it would be different).

I've opened issue #73 to discuss this.

PR #23, opened 2016-04-11
Fix for parsing "null" values for auto-value objects.

radzio:

Issue #67, opened 2016-09-14
Unable to Serialize null field


dkpalmer:
I have a need to serialize a null field. When I mark the field as `@Nullable`, I see the GsonTypeAdapter `write(JsonWriter, Object)` method wrap the field adapter's `write()` call in a null check. But if I remove `@Nullable`, my builder/factory methods complain about the null value. Is there any way around this?

It looks like [23](https://github.com/rharter/auto-value-gson/pull/23) could be related?

For clarity, I am calling serializeNulls in my GsonBuilder. Perhaps the write method's GsonTypeAdapter could check that flag in the JsonWriter to bypass the above mentioned null check?

hzsweers:
Closed by #78

Issue #50, opened 2016-06-10
Generated TypeAdapterFactory and null safety


muddin1:
As of 0.2.5, My custom AutoValueTypeAdapterFactory would call `nullSafe()` on every type adapter I return in the `create()` method. Refer https://github.com/rharter/auto-value-gson/issues/28#issuecomment-211680677.

``` java
Class<? super T> rawType = type.getRawType();
if (rawType.equals(User.class)) {
    return (TypeAdapter<T>) User.typeAdapter(gson).nullSafe();
}
```

Beginning from 0.3.0, the generated AutoValueGsonTypeAdapterFactory does not call `nullSafe()` by default, which is the appropriate way for sure.

So, how'd I go about ensuring null-safety when using the generated TypeAdapterFactory.

rharter:
You could submit a PR if you've got the code.  Otherwise I'll have to add that.

muddin1:
Are you saying that we should attach `nullSafe()` to every AutoValue typeAdapter mentioned in the generated Factory i.e. in the annotation processor?

Also, there's this PR https://github.com/rharter/auto-value-gson/pull/23 that presumably eliminates the need to use `nullSafe()` completely.

jachenry:
Below is a quick solution for anyone else that hits this before #23 is completed.

```java
@GsonTypeAdapterFactory
public abstract class AutoValueAdapterFactory implements TypeAdapterFactory {

  public static TypeAdapterFactory create() {
    final TypeAdapterFactory factory = new AutoValueGson_AutoValueAdapterFactory();

    return new TypeAdapterFactory() {
      @Override
      public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
        TypeAdapter<T> typeAdapter = factory.create(gson, type);
        return typeAdapter != null ? typeAdapter.nullSafe() : null;
      }
    };
  }
}
```

muddin1:
@jachenry Thanks, I've been doing the same.

jjimenez0611:
Hi @jachenry 

Please can you explain to me with more details..

I have my factory like this:

public class **AutoValueGsonTyeAdapterFactory** implements TypeAdapterFactory  {

    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
        Class<? super T> rawType = type.getRawType();

        if (rawType.equals(MGuest.class)) {
            return (TypeAdapter<T>) MGuest.typeAdapter(gson);
        }

[...]

PR #78, opened 2016-10-24
Allow writing and reading null values

ansman:
This closes #23, #50 and #67


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -336,20 +336,18 @@ public MethodSpec createWriteMethod(TypeName autoValueClassName,
...
+    writeMethod.addStatement("$N.nullValue()", jsonWriter);
+    writeMethod.addStatement("return");
+    writeMethod.endControlFlow();
+
     writeMethod.addStatement("$N.beginObject()", jsonWriter);
     for (Map.Entry<Property, FieldSpec> entry : adapters.entrySet()) {
       Property prop = entry.getKey();
       FieldSpec field = entry.getValue();
 
-      if (prop.nullable()) {
```
ansman: I considered using `nullValue()` instead of letting the adapter handle it but adapters should handle `null` values according to the documentation:
> Type adapters should be prepared to read null from the stream and write it to the stream. Alternatively, they should use nullSafe() method while registering the type adapter with Gson.

rharter:
LGTM 👍

PR #83, opened 2016-11-01
Add support for custom field adapters

### LINKED ISSUE #82: Implement support for per field TypeAdapters
Gson has a `@JsonAdapter` annotation that allows you to specify a custom type adapter for just that field, that would be awesome to support :)

ansman:
This closes #82

### DIFF
--- a/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -41,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.annotation.processing.Messager;
+import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
@@ -51,6 +52,7 @@ import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
+import javax.lang.model.util.Types;
 import javax.tools.Diagnostic;
 
 import static javax.lang.model.element.Modifier.ABSTRACT;
@@ -68,7 +70,7 @@ public class AutoValueGsonExtension extends AutoValueExtension {
     final ExecutableElement element;
     final TypeName type;
     final ImmutableSet<String> annotations;
-    final TypeName typeAdapter;
+    final TypeMirror typeAdapter;
 
     public Property(String humanName, ExecutableElement element) {
       this.methodName = element.getSimpleName().toString();
@@ -78,12 +80,7 @@ public class AutoValueGsonExtension extends AutoValueExtension {
       type = TypeName.get(element.getReturnType());
       annotations = buildAnnotations(element);
 
-      TypeMirror typeAdapterAnnotation = getAnnotationValue(element, GsonTypeAdapter.class);
-      if (typeAdapterAnnotation == null) {
-        typeAdapter = null;
-      } else {
-        typeAdapter = TypeName.get(typeAdapterAnnotation);
-      }
+      typeAdapter = getAnnotationValue(element, GsonTypeAdapter.class);
     }
 
     public static TypeMirror getAnnotationValue(Element foo, Class<?> annotation) {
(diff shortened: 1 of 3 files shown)


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -73,6 +76,42 @@ public Property(String humanName, ExecutableElement element) {
...
+
+      TypeMirror typeAdapterAnnotation = getAnnotationValue(element, GsonTypeAdapter.class);
+      if (typeAdapterAnnotation == null) {
+        typeAdapter = null;
+      } else {
+        typeAdapter = TypeName.get(typeAdapterAnnotation);
+      }
+    }
+
+    public static TypeMirror getAnnotationValue(Element foo, Class<?> annotation) {
```
ansman: It seems like this is the only way of getting the annotation value without catching exceptions, but perhaps there is a better way.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -293,7 +332,9 @@ public TypeSpec createTypeAdapter(ClassName className, ClassName autoValueClassN
     for (Map.Entry<Property, FieldSpec> entry : nonparameterizedAdapters.entrySet()) {
       Property prop = entry.getKey();
       FieldSpec field = entry.getValue();
-      if (entry.getKey().type instanceof ParameterizedTypeName) {
+      if (prop.typeAdapter != null) {
+        constructor.addStatement("this.$N = new $T()", field, prop.typeAdapter);
```
ansman: Should we perhaps cast the class to the adapter field's type?

JakeWharton:
Hesitant about this because it doesn't support `TypeAdapterFactory` or `JsonSerializer` / `JsonDeserializer` which are also methods of providing custom conversion. I don't really care about the latter, but `TypeAdapterFactory` is essential for writing some `TypeAdapters` since they are not otherwise provided the `Gson` instance for delegation.

ansman:
@JakeWharton I considered adding support for factories and it makes sense to add it. Will do that tomorrow.

ansman:
@JakeWharton I've added support for factories now.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -289,16 +325,48 @@ public TypeSpec createTypeAdapter(ClassName className, ClassName autoValueClassN
...
       Property prop = entry.getKey();
       FieldSpec field = entry.getValue();
-      if (entry.getKey().type instanceof ParameterizedTypeName) {
-        constructor.addStatement("this.$N = $N.getAdapter($L)", field, gsonParam,
-            makeType((ParameterizedTypeName) prop.type));
+      if (prop.typeAdapter != null) {
+        if (typeUtils.isAssignable(prop.typeAdapter, typeAdapterFactory)) {
+          constructor.addStatement("this.$N = new $T().create($N, $L)", field, TypeName.get(prop.typeAdapter),
+                  gsonParam, makeType(prop.type));
+        } else {
```
ansman: Should we verify that the typeAdapter is actually a `TypeAdapter` and abort the processing if it isn't?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -289,13 +325,27 @@ public TypeSpec createTypeAdapter(ClassName className, ClassName autoValueClassN
...
+
     ImmutableMap<Property, FieldSpec> nonparameterizedAdapters = filterParameterizedAdapters(adapters, false);
     for (Map.Entry<Property, FieldSpec> entry : nonparameterizedAdapters.entrySet()) {
       Property prop = entry.getKey();
       FieldSpec field = entry.getValue();
-      if (entry.getKey().type instanceof ParameterizedTypeName) {
+      if (prop.typeAdapter != null) {
+        if (typeUtils.isAssignable(prop.typeAdapter, typeAdapterFactory)) {
+          constructor.addStatement("this.$N = new $T().create($N, $L)", field, TypeName.get(prop.typeAdapter),
+                  gsonParam, makeType(prop.type));
```
ansman: Should we use `TypeToken#get(Class)` instead of `new TypeToken<Class>(){}` for non parameterized types?

JakeWharton: Yes. It saves a lot of runtime reflection overhead.

rharter:
LGTM.  Fix up the test and it's ready to merge.

ansman:
@rharter I've fixed the tests and rebased with master. But what about my outstanding comments, any we should take a look at?

rharter:
Regarding verifying that what the user puts in `@GsonTypeAdapter` is, in fact, a `TypeAdapter`?   Seems unnecessary to me, but I'm open to other thoughts.

ansman:
Yeah, it won't compile so that's fine.

ansman:
I've rebased against master now

PR #84, opened 2016-11-02
Implement support for default values

ansman:
Usage:
```java
public static TypeAdapter<Person> typeAdapter(Gson gson) {
    return new AutoValue_Person.GsonTypeAdapter(gson)
        .setDefaultName("Bruce Wayne");
}
```


ansman:
Another suggestion is to use builders instead, that would take some more work though.

rharter:
I like this approach, since builders are optional, I think this makes sense.

plaflamme:
@rharter Any chance we can get a `0.4.5` with this change? Also snapshot dependencies don't seem to be published currently: https://oss.sonatype.org/content/repositories/snapshots/com/ryanharter/auto/value/auto-value-gson/

PR #88, opened 2016-11-08
Update gradle to 3.1.

serj-lotutovici:


3flex:
Consider updating to the newer Gradle 3.2.1 instead

rharter:
Thanks for this, but since it took me so long I just upgraded to 3.2.1.

PR #89, opened 2016-11-08
(demo/proposal) Support default values on methods via annotations

hzsweers:
This is to demonstrate an API of how default values could be applied directly onto methods via annotation. This is a nicer API in my opinion, but both it and the current setter approach in #84 (CC @ansman) have their pros and cons. I reverted it in this PR so implementations wouldn't conflict. Tests will fail though, will update if we proceed with this.

This approach:
  - Pros:
      - Adapters remain immutable and not stateful
      - Simple, very easy to use API. Default behavior is more obvious to the reader with its in-code locality to the field
      - `@DefaultValue` pattern could be standard across auto-value-gson/moshi/parcel. 
          - This would also open the door to possibly sharing this and other future annotations (`@Transient` has been discussed before) between them for the same behavior.
      - Simple to maintain/support under the hood as well
      - Default value becomes opt-in
  - Cons:
      - Only works for builtin types unless we want to support dropping in json strings
      - Not type-safe (though it will fail at compile-time)
         - Disclaimer: this is less of an issue to me since we generate all our models, but figured it probably falls under "con" :)
      - Requires an extra component (the annotation)

 Setter approach:
   - Pros:
       - Simple to use
       - Supports any type
       - Typ-safe
       - No extra components needed, all generated from the current API
   - Cons:
       - Adapters become mutable and stateful
       - Pattern would only work for auto-value-gson/moshi, not parcel
       - Not obvious if/what a field's default is - behavior is defined by the adapter
       - No opt-out mechanism. All fields can have default values


hzsweers:
So code like this:

screen shot 2016-11-08 at 12 14 57 pm

Would look like this:

screen shot 2016-11-08 at 12 14 42 pm

serj-lotutovici:
I defiantly like the approach of sharing annotations between the extension libraries as a standard of some sorts. This could potentially create a very fruitful environment for ideas which would make auto-value-gson/moshi/parce/<others> thrive from one another. 

Nice job @hzsweers 👍

ansman:
I feel like not being able to support non built in types is a deal breaker.
Otherwise I really like this approach.

ansman:
Also, even though the approach using setters means it technically has state it's not a big problem i real life since you usually don't keep a reference to type adapter yourself.

rharter:
I've been playing around with this tonight, and I agree that the limitation of only built in types is a big challenge.  I was looking into basically using the default value string as an initialization statement, but that doesn't cover imports, so there's challenges with that as well.

hzsweers:
What about having both with complex types forcing the creation of a builder?

`@DefaultValue` to opt-in to defaults. You can supply a string or you can point it to some string constant of it called `DELEGATE`, which would be a signal to the generator to create a builder option for your typeadapter. 

Specifically: 

If there's any `@DefaultValue(DELEGATE)`, then a builder option is generated. I'd suggest matching Dagger in this way, with a static `builder()` method that returns a builder. For every delegating default, you add an option to the builder, returning an inner `DelegateBuilder` that either constructs a private inner `Delegate` with getters or you can make a private constructor that accepts all the parameters (like how AV does). Then the actual TypeAdapter itself is still stateless, you get opt-in, and you can stick with simple defaults if you want.

Let me know if that makes sense. Obviously would need a little more ironing but I think this gets wins for everyone. It also still stays agnostic to AV-gson, so you can still share this.

ansman:
Supporting builders would preferred in every way IMO.

hzsweers:
Talked with @ansman separately. I think relying on AV's generated builders, though ideal (wouldn't need to specify defaults in AVG at all), isn't really feasible without support in AV's extensions API. Between `is/set` prefixes, `Optional` support, and general tedium with trying to manually parse the builder API ourselves, I think this is the best recourse for now.

ansman:
I feel like the `@DefaultValue(DELEGATE)` feels very clunky, there should be a more explicit way to do this.

hzsweers:
Any suggestions? Maybe a separate `@DelegateDefault` annotation?

ansman:
Perhaps, or you could simply annotate the `typeAdapter` method to indicate that you want a builder?

hzsweers:
How would you indicate which fields to include in the builder?

ansman:
Couldn't all fields be generated?

hzsweers:
I'd much prefer they be opt-in. Would rather be limited by default and flexible than eager and wasteful. In our case, we have over 1000 models that all don't need a builder

ansman:
Sure, the builder doesn't always have to be generated but a method for each field doesn't seem wrong to me.

hzsweers:
@ansman Let's try to find a path forward here. I've opened an issue on autovalue about getting Builder information here - https://github.com/google/auto/issues/421. In the meantime, I don't think we should really block on that, as it's out of my hands and a little unfair to disregard other proposals on it since the current one doesn't leverage real builders either. 

Another idea I thought of is something a little a la Dagger, where the consumer can define their own TypeAdapter builder interface. It's more work on the code gen part, but gives you sort of ultimate flexibility. 

Given

```java
@AutoValue
abstract class Thing {
 @DefaultValue("2")
 public abstract int someInt();

 public abstract AnotherThing anotherThing();

  @AutoValueGson.Builder
  public interface TypeAdapterBuilder {
    TypeAdapterBuilder anotherThing(AnotherThing thing);
  }
}
```

Delegate builder

```java
static class GsonTypeAdapter {

 private final int defaultSomeThing;
 private final AnotherThing defaultAnotherThing;

 GsonTypeAdapter() {
   this(2, null);
 }

 private GsonTypeAdapter(int someInt, AnotherThing anotherThing) {
   // Assign them
 }

 TypeAdapterBuilder builder() {
    return new DelegateBuilder();
 }

 // Later down
 static class DelegateBuilder implements TypeAdapterBuilder {
   private int someThing = 2;
   private AnotherThing anotherThing;

   // Fill out setters, but not one for someThing
   @Override
   public TypeAdapterBuilder anotherThing(AnotherThing thing) {
     this.anotherThing = thing;
     return this;
   }

   // Finally
   TypeAdapter build() {
     return new GsonTypeAdapter(someThing, anotherThing);
   }
 }
}
```

Finally

```java
// Complex objects are still possible
public static TypeAdapter typeAdapter(Gson gson) {
   return AutoValue_Thing.GsonTypeAdapter.builder()
       .anotherThing(/** impl here */)
       .build();
}

// So do simple ones
public static TypeAdapter typeAdapter(Gson gson) {
   return new AutoValue_Thing.GsonTypeAdapter();
}
```

I propose we go with this or what I proposed earlier with the opt-in annotation. If AV adds some builder APIs in the future, then great we can leverage them. In the meantime I'd like to get somewhere with this.

@rharter any thoughts?

rharter:
Closing as this is super old.

PR #91, opened 2016-11-17
Allow ignoring fields for serialization/deserialization

ajarl:
Supports ignoring serialization or deserialization (or both). If both are ignored, no adapter for that type will be generated, but you can still set a default value to use if you want.

This is intended to basically be a replacement for `transient` fields.


3flex:
I've been looking into this recently as well. Just wondering if you've tried Gson's [ExclusionStrategy](https://google.github.io/gson/apidocs/com/google/gson/ExclusionStrategy.html) to achieve this?

Just curious if you've tried this and found it lacking in some way?

ajarl:
@3flex: As far as I can tell, the ExclusionStrategy is only ever used by the ReflectiveTypeAdapterFactory. Since the point of auto-value-gson is to avoid the reflective type adapters, the ExclusionStrategy won't ever be applied.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -809,4 +812,121 @@ public void simple() {
...
+        + "            c = cAdapter.read(jsonReader);\n"
+        + "            break;\n"
+        + "          }\n"
+        + "          default: {\n"
+        + "            jsonReader.skipValue();\n"
+        + "          }\n"
+        + "        }\n"
+        + "      }\n"
+        + "      jsonReader.endObject();\n"
+        + "      return new AutoValue_Test(a, b, c, d);\n"
```
rharter: This results in code that will fail to compile when integrated with AutoValue.  The issue here is that `d` will always be null, but wasn't annotated `@Nullable`, so it's illegal to pass null for it to the AutoValue constructor.

This means that fields set to ignore deserialization **must** also me marked `@Nullable`.  For serialization, it doesn't really matter.

ajarl: I don't think that's true. First, it won't fail to compile, since AutoValue allows you to pass null (it will throw in runtime though). Second, the value of `d` here is not always null, it's actually whatever you've set as default for that field with `setDefaultD`. We obviously can't tell if the default value has been set or not during codegen, so I don't think there's a good way to handle that here.

Basically, the effect is that if you mark something as ignored for deserialization *and* you haven't annotated it `@Nullable` *and* you haven't set a non-null default value for said field on the generated adapter, deserialization will fail in runtime with an NPE. To me, this is expected behavior.

The alternative is to remove the concept of default values specifically for fields that are ignored by deserialization, and enforce them to deserialize as null - then we can enforce that said property is marked as nullable during codegen. This seems unnecessarily restrictive though, IMO.

If I misunderstood something, let me know :)

michaelcarrano:
Any updates on this? Would be nice to be able to ignore fields.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
rharter: Hmm, the challenge I'm having with this is that I'd like to avoid a situation where we generate code that is easy to use in a breaking fashion.  In this case, the user consuming this generated TypeAdapter has no way of knowing which default items are required or not, so it's pretty easy to end up in a situation in which your code compiles, runs fine during testing, then a minor change in the JSON response from a server 6 months later will break the code in the wild.  That scares me.

You're right that we can't tell if the default value has been set at every usage during code gen, but we have the other pieces that we can use to make that an unreachable case.  For instance, in the case where an item is ignored for deserialization **and** it's not `@Nullable`, we know this property requires a default value.  Therefore, we could simply add that as a parameter to the constructor of the type adapter, instead of making it a property.  This makes it required and documented (by code), and also means that you can't actually construct a TypeAdapter in a broken state, as opposed to the current code in which `new TestTypeAdapter(gson)` gives you a non-valid (for some responses) object.

Perhaps even logging a warning or info message during generation will help users understand **why** this is happening.

I think this is a good solution, just want to get past that last piece to help users out.

ajarl: I agree, this is good input. I'll take another look at this as soon as I have time. Thanks for the comments.

ajarl:
@rharter I rebased and fixed so you now have to supply default values for non-nullable fields that are ignored on deserialization to the constructor of the TypeAdapter. You can of course still supply null there, not sure if it's worth adding verification of it there, since presumably the implementor is aware that said property is not nullable.

But let me know if anything still bothers you :)

rharter:
This update looks good, but I have one question that I'd like some feedback on.

The `@Ignore(serialization = false)` seems like it can be a little confusing, so I'd like to know thoughts.  On first glance, it's hard to understand that we're basically saying serialization is not not ignored.  Is there a better way we can phrase this?

/cc @hzsweers any thoughts?

ajarl:
Yeah, I had the same concerns, discussed it with @ansman a little bit. The use case we decided to optimize for was that where you want to simply ignore the field entirely (à la `transient`), so just adding an `@Ignore` would be the nicest way of doing that. If the partial serialization was the more common use case, one might want to see a `@SerializationStrategy(serialize = false)` or something, but it means it's more verbose and/or confusing for the case where you want to entirely ignore a field.

A third idea that may be better is to just introduce some kind of enum or int type that the `Ignore` annotation would take as a parameter, like `@Ignore` to ignore entirely, `@Ignore(Ignore.SERIALIZATION)` to ignore serialization (but not deserialization), etc.

rharter:
That second idea makes more semantic sense, I think. It could be really clear by defaulting the value to `BOTH`, and if you want to only ignore one you write `@Ignore(DESERIALIZATION)`, which matches the conversational directive.

I think the challenge with the current solution is just that it looks very similar to `@Ignore(serialize = false)`, which would mean the opposite of `@Ignore(serialization = false)`.  It's a question of if the boolean pertains to serialization, or ignoring serialization.

Would you mind making that quick update then I'll merge this?

ajarl:
@rharter: Fixed :)

naturalwarren:
Nice work!

eliasbagley:
Is there a release that has this in it?

thevery:
@eliasbagley looks like not yet, but you can use jitpack for it:
```
repositories {
    maven { url 'https://jitpack.io' }
}
```
    annotationProcessor 'com.github.rharter:auto-value-gson:4af01db'
    provided 'com.github.rharter:auto-value-gson:4af01db'

eliasbagley:
Perfect that works for now.

PR #99, opened 2016-12-23
Use TypeToken.getParameterized for all parameterized fields

benkay:
This eliminates the remaining reflection when creating TypeAdapters.

In a test app on a Galaxy J1, it reduces the time taken to initialize 10 `TypeAdapter`s for classes each containing 10 `List<String>` fields from 750 - 800ms to ~30ms

I also enabled custom field adapter support for parameterized generic fields. Not sure it's useful, but it simplified the code.


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -330,7 +330,6 @@ public TypeSpec createTypeAdapter(Context context, ClassName className, ClassNam
             .addParameter(gsonParam);
 
     if (!typeParams.isEmpty()) {
-      ImmutableMap<Property, FieldSpec> parameterizedAdapters = filterParameterizedAdapters(adapters, true);
```
JakeWharton: nit: delete empty line

thevery:
Hm, looks like this doesn't work with proguard nicely:

Warning: com.yandex.mail.react.entity.$AutoValue_Fields$GsonTypeAdapter: can't find referenced method 'com.google.gson.reflect.TypeToken getParameterized(java.lang.reflect.Type,java.lang.reflect.Type[])' in program class com.google.gson.reflect.TypeToken

thevery:
upd: not an auto-value-gson problem but rather gson: TypeToken.getParameterized(List.class, Recipient.class) was added in gson 2.8.0

thevery:
upd2: my bad, forgot to update explicit gson dependency to 2.8.

PR #101, opened 2016-12-30
Serialize collections to null by default, add compiler option to serialize to empty variants

### LINKED ISSUE #73: What should the default collection policy be?
Recently the functionality of collections was changed to default to empty collections, instead of null (#71), but this technically breaks the contract with AutoValue, since the property is not annotation with `@Nullable`.  That means that it should technically exist in the JSON.  This potentially masks JSON encoding issues.

IMHO, the issue here is that deserializing and serializing are now no longer compatible, i.e. if you deserialize JSON that has a null collection property then serialize it using this extension, you won't be able to reproduce the original JSON, potentially with different meaning.

Recognizing the usefulness of this functionality, @hzsweers has proposed making this a non-default option configurable using apt arguments.  This ticket is to discuss the implications and determine what the default collection policy should be.

hzsweers:
Resolves #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,20 @@ Gson gson = new GsonBuilder()
     .create();
 ```
 
+## Compiler options
+
+`autovaluegson.defaultCollectionsToEmpty` - If specified, maps/collections will default to their empty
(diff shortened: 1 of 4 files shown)


rharter:
retest

rharter:
I've fixed the build failures in #103

PR #108, opened 2017-01-19
Update sample with  @GsonTypeAdapter

kmenager:
Add sample for custom @GsonTypeAdapter annotation

### DIFF
--- a/example/src/main/java/com/ryanharter/auto/value/gson/example/BirthdateAdapter.java
+++ b/example/src/main/java/com/ryanharter/auto/value/gson/example/BirthdateAdapter.java
@@ -32,13 +32,7 @@ public class BirthdateAdapter extends TypeAdapter<Date> {
     try {
       return df.parse(stringDate);
     } catch (ParseException e) {
-      throw new DateTypeParseException(stringDate);
-    }
-  }
-
-  public class DateTypeParseException extends RuntimeException {
-    public DateTypeParseException(String stringDate) {
-      super("Unparseable date: " + stringDate + ". It must be in \"yyyy-MM-dd\" format");
+      return new Date();
     }
   }
 }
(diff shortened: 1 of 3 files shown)


rharter:
Looks like the `PersonTest` will also need to be updated.

kmenager:
@rharter Yeah, I just fix it

example/src/main/java/com/ryanharter/auto/value/gson/example/BirthdateAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+  @Override public Date read(JsonReader in) throws IOException {
+    if (in.peek() == JsonToken.NULL) {
+      in.nextNull();
+      return null;
+    }
+    String stringDate = in.nextString();
+    try {
+      return df.parse(stringDate);
+    } catch (ParseException e) {
+      throw new DateTypeParseException(stringDate);
```
rharter: I'd much rather not see a runtime exception in the sample, as in the real world this will cause the app to crash, and that's not what we want to encourage.  Instead, perhaps return a default value like `new Date()`

example/src/test/java/com/ryanharter/auto/value/gson/example/PersonTest.java
```diff
@@ -2,22 +2,33 @@
...
 
 public class PersonTest {
     @Test
     public void testGson() throws Exception {
+        final DateFormat df = new SimpleDateFormat("yyyy-MM-dd", Locale.getDefault());
+
+        String birthdate = "2007-11-11";
+        Date date = df.parse(birthdate);
+
+
```
rharter: nit: remove extra whitespace

kmenager:
@rharter Is my modification ok ?

rharter:
Thanks!

PR #116, opened 2017-02-17
set auto-service dependency to compileOnly

gabrielittner:
It's only needed at compile time to generate the service file.


hzsweers:
Thanks!

auto-value-gson/build.gradle
```diff
@@ -8,7 +8,7 @@ dependencies {
     compile 'com.google.code.gson:gson:2.8.0'
     compile 'com.squareup:javapoet:1.8.0'
     compile 'com.google.auto.value:auto-value:1.3'
-    compile 'com.google.auto.service:auto-service:1.0-rc2'
+    compileOnly 'com.google.auto.service:auto-service:1.0-rc2'
```
naturalwarren: rc3 is out if you'd like to pick that up :)

PR #117, opened 2017-02-21
Fixes #110 - Support inner Factory class generation

### LINKED ISSUE #110: Factory class can't be inner class?
If I declare the factory class inside another class, the generated factory class name will contains `.`, which is illegal, could it be substituted with `_`?

R4md4c:
I guess the problem was on how AutoValueGson was generating the class name, I checked how AutoValue generates the class name[1] and came up with this solution, hope that you'd like it, if not then I'm all ears for improvements. 

Also If you'd like, I could add an example code too in the 'example' project.

[1] https://github.com/google/auto/blob/master/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java#L216


R4md4c:
Hi @rharter, Any chance fo reviewing this PR?

rharter:
Sorry for the delay, this looks great! Thanks!

PR #124, opened 2017-03-31
Respect FieldNamingStrategy when set on GsonBuilder

### LINKED ISSUE #18: Add support for field naming policies
Gson has support for different field naming policies which does not seem to work with this library. For example, by setting LOWER_CASE_WITH_UNDERSCORES while creating the gson object, all field names with lowercase and underscore separated format will automatically mapped to their camel case formats in the Java classes.

```
Gson gson = new GsonBuilder()
    .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
    .create();

public class Dog {
    public String makeNoise;
}
```

**And its equivalent json:**

```
{
    "make_noise": "bark"
}
```

We can use @SerializedName annotation for this but it becomes tedious if all the field names are to be converted from the same underscore format.

3flex:
Closes #18.

The code generated by the `createFieldNamingMethods()` method is taken from the Gson project, so not sure how to attribute this to them; it was derived from their code but is now generated with JavaPoet. I left their license header on `FieldNamingTest.java` because that only has slight modifications from the code taken from their project.

This will require projects using auto-value-gson to have a runtime dependency on Gson 2.8.4+ going forward.


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -547,6 +572,94 @@ public MethodSpec createReadMethod(ClassName className,
...
+            .beginControlFlow("if (!$T.isUpperCase(firstCharacter))", Character.class)
+            .addStatement("$T modifiedTarget = modifyString($T.toUpperCase(firstCharacter), $N, ++index)", String.class, Character.class, name)
+            .addStatement("return fieldNameBuilder.append(modifiedTarget).toString()")
+            .nextControlFlow("else")
+            .addStatement("return name")
+            .endControlFlow()
+            .build();
+
+    MethodSpec renameFieldMethod = MethodSpec.methodBuilder("renameField")
+            .addModifiers(PUBLIC, STATIC)
```
3flex: Should be private, not public

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -547,6 +572,94 @@ public MethodSpec createReadMethod(ClassName className,
...
+    ClassName fieldNamingPolicy = ClassName.get("com.google.gson", "FieldNamingPolicy");
+    ClassName locale = ClassName.get("java.util", "Locale");
+
+    MethodSpec modifyStringMethod = MethodSpec.methodBuilder("modifyString")
+            .addModifiers(PRIVATE, STATIC)
+            .returns(String.class)
+            .addParameter(firstCharacter)
+            .addParameter(srcString)
+            .addParameter(indexOfSubstring)
+            .addStatement("return (indexOfSubstring < srcString.length()) ? firstCharacter + srcString.substring(indexOfSubstring) : String.valueOf(firstCharacter)")
```
3flex: Use $N substitution instead of hardcoding field references

example/src/test/java/com/ryanharter/auto/value/gson/example/FieldNamingTest.java
```diff
@@ -0,0 +1,91 @@
...
+                .UpperCamel(2)
+                ._lowerCamelLeadingUnderscore(3)
+                ._UpperCamelLeadingUnderscore(4)
+                .lower_words(5)
+                .UPPER_WORDS(6)
+                .annotated(7)
+                .lowerId(8)
+                .build();
+    }
+}
```
3flex: add newline

3flex:
Added notes to correct a couple of minor issues I just noticed. I'll fix them when incorporating any feedback people have.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -404,10 +410,12 @@ public TypeSpec createTypeAdapter(Context context, ClassName className, ClassNam
         .superclass(superClass)
         .addFields(adapters.values())
         .addFields(defaultValueFields.values())
+        .addField(FieldSpec.builder(Gson.class, "gson", PRIVATE).build())
```
rharter: As opposed to storing a reference to the Gson instance, it would be cleaner to just store the result of `fieldNamingStrategy()` here.  That's all it's used for in the `renameField()` method, and storing that would reduce the number of method call in the adapter, since we'd have a cached value.  

The FieldNamingStrategy is immutable, so it won't change at runtime unless the user fetches a new instance of Gson, thereby creating a new TypeAdapter.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -547,6 +572,94 @@ public MethodSpec createReadMethod(ClassName className,
...
+            .endControlFlow()
+            .build();
+
+    MethodSpec renameFieldMethod = MethodSpec.methodBuilder("renameField")
+            .addModifiers(PUBLIC, STATIC)
+            .returns(String.class)
+            .addParameter(name)
+            .addParameter(gson)
+            .addException(IOException.class)
+            .beginControlFlow("if ($N.fieldNamingStrategy().equals($T.UPPER_CAMEL_CASE))", gson, fieldNamingPolicy)
```
rharter: This would be cleaner as a switch case, and reduce the number of calls to `fieldNamingStrategy()`.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -508,19 +520,32 @@ public MethodSpec createReadMethod(ClassName className,
...
 
     // skip value if field is not serialized...
     readMethod.beginControlFlow("default:");
+    for (Property prop : properties) {
+      if (!prop.shouldDeserialize()) {
+        continue;
+      }
+      if (!prop.hasSerializedNameAnnotation()) {
+        FieldSpec field = fields.get(prop);
+        readMethod.beginControlFlow("if (renameField($S, gson).equals(_name))", prop.humanName);
```
rharter: Since we're in the business of talking about caching, and the FieldNamingStrategy is immutable per TypeAdapter instance, instead of calling `renameField` on every serialize/deserialize pass, the renamed variant can be cached in a member variable on the TypeAdapter to speed up serialization.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
3flex: FieldNamingStrategy is an interface, but this PR only supports FieldNamingPolicy, an enum which implements FieldNamingStrategy. I could use the switch statement after first checking if FieldNamingStrategy is an instance of FieldNamingPolicy, would that be appropriate?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
3flex: Would it be OK to do a single pass when the TypeAdapter is instantiated, store the renamed field names in a map, then reference those when seralizing/deserializing? I know there's been some previous work to lazy load elements within the TypeAdapter to improve performance so not sure if doing this when the class is created would also be considered too slow.

3flex:
Finally getting around to finishing this off. I haven't pushed my changes yet, but I've rebased on master and also I've addressed the feedback with the exception of a couple of things that I've mentioned in replies to the original review feedback. Once those are fixed up I'll push a new update for review.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
rharter: That works for me.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
rharter: Makes sense.  While we can skip reflection in the default case, I wonder if it makes sense to support this with reflection for the case where the user uses a non-default naming policy. The reflection code will never be hit unless the user explicitly chooses to use something that requires reflection, i.e. a custom naming policy based on `Field`, making the perf loss a conscious choice.

```java
FieldNamingStrategy strategy = gson.fieldNamingStrategy();
if (strategy instanceof FieldNamingPolicy) {
  switch ((FieldNamingPolicy) strategy) {
    ...
  }
} else {
  return strategy.translateName(MyClass.class.getField("foo"));
}
```

3flex:
Updated with feedback taken into account; let me know if there's anything further that needs addressing.

3flex:
@rharter just noticed you released a new version a few days ago without merging this PR - any feedback you'd like me to address?

3flex:
@rharter sorry to bother you - just wondering if you have further thoughts or feedback on this PR? If I have to rebase again I probably wouldn't go through the trouble, so if you think there's value here it'd be best to merge before other changes go in.

If not feel free to close, cheers.

rharter:
Sorry for letting this drop off my radar, @3flex.  Thanks for the contribution!

Issue #118, opened 2017-02-26
Add a separate artifact for the annotations


ansman:
It would be awesome if there was a separate artifact for the annotations to avoid having the whole auto value codebase in your classpath when compiling.

jonathan-caryl:
+1 for this

anthonymonori:
This should be linked with #126

hzsweers:
This is done!

PR #126, opened 2017-04-06
Standalone annotations artifact (#118)

lgawin:
I ~used JakeWharton's annotations for AutoValue (https://github.com/JakeWharton/AutoValueAnnotations) and~ created separate module with annotations only. And now it is possible to use `auto-value-gson` extension with the latest AS (2.4.0-alpha4) by specifying following dependencies:

```
    annotationProcessor "com.ryanharter.auto.value:auto-value-gson:$latestVersion"
    provided "com.ryanharter.auto.value:auto-value-gson-annotations:$latestVersion"
```


auto-value-gson/build.gradle
```diff
@@ -5,16 +5,19 @@ sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
 
 dependencies {
-    compile 'com.google.code.gson:gson:2.8.0'
+    compile project(':auto-value-gson-annotations')
+    compile deps.gson
     compile 'com.squareup:javapoet:1.8.0'
-    compile 'com.google.auto.value:auto-value:1.3'
+    compile 'com.jakewharton.auto.value:auto-value-annotations:1.4'
+    compileOnly 'com.google.auto.value:auto-value:1.4'
```
JakeWharton: This isn't right. This module should have a compile dependency on auto value because it requires it for runtime (read: processing time) behavior.

lgawin: So I don't need to use your annotations here and can just use
```
compile 'com.google.auto.value:auto-value:1.4'
```
Right?

lgawin: Sure, this artifact is then used as `annotationProcessor` dependency later... 
I think `compileOnly` for `auto-service` made me wrong... Thanks for the comment.

anthonymonori:
Is there anything that's missing here?

PR #133, opened 2017-06-12
Make defaults off by default and some dependency updates

hzsweers:
This PR disables the defaults handling by default, as it imposes quite a bit of bloat (one method per property) onto users that may not actually need/want it.

This also opportunistically updates dependencies and to the new java-library format


hzsweers:
Sidenote - updating gradle's helped me get around all the weird local build issues I'd had on previous PRs.

build.gradle
```diff
@@ -20,3 +20,8 @@ allprojects {
         mavenCentral()
     }
 }
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '4.0-rc-2'
```
serj-lotutovici: Gradle 4.0 is already released.

bejibx:
But how to enable defaults generation now?

hzsweers:
Use the compiler option

Issue #132, opened 2017-06-07
Request: Support property normalization and validation


bryanstern:
Currently the generated `GsonTypeAdapter` uses the generated constructor when creating an instance of the class being deserialized. Unfortunately, this skips over any additional normalization or validation the class might doing in a factory method or builder.

It'd be great the generated `GsonTypeAdapter` accepted an optional custom factory. Something like:
```java
package com.ryanharter.auto.value.gson.example;

import com.google.auto.value.AutoValue;
import com.google.common.base.Preconditions;
import com.google.gson.Gson;
import com.google.gson.TypeAdapter;

@AutoValue public abstract class User {
  abstract String firstname();
  abstract String lastname();
  abstract int age();

  public static User create(String firstName, String lastName, int age) {
    Preconditions.checkArgument(!firstName.isEmpty());
    Preconditions.checkArgument(!lastName.isEmpty());
    if (age < 0) {
      age = 0;
    }

    return new AutoValue_User(firstName, lastName, age);
  }

  public static TypeAdapter<User> typeAdapter(Gson gson) {
    return new AutoValue_User.GsonTypeAdapter(gson, new AutoValue_User.GsonTypeAdapter.Factory() {
      @Override
      public User create(String firstname, String lastname, int age) {
        return User.create(firstname, lastname, age);
      }
    });
  }
}
```

Thoughts?

hzsweers:
there's currently some default value support that we've planned to turn off by default in favor of pushing for access to autovalue builders in the extensions API, but this is an interesting alternative since it only makes it one added method rather than a bunch of mutating ones that the generated adapters currently have.

Do you have any other examples? Of the one you sent, my first inclination would actually be to just drop that model on the floor as invalid, or at least move the validation out.

bryanstern:
I agree that example is kind of trivial. The main benefit of this is that it allows the same validation/normalization guarantees whether the object was created via a factory method, builder, or was deserialized from json.

My specific motivation is that I am receiving some json I need to deserialize and unfortunately the schema is not well respected by the sender resulting in writing custom deserializers to handle this.

One example that comes to mind is where two properties might be optional, but they should either both be present or absent. There isn't a great way to correct for this unless.

rharter:
This would also be good to support without Builders, like in your example, and could probably already be done by simply annotating the main, all arg static factory method that you want the TypeAdapters to use.

[...]

PR #136, opened 2017-07-06
Support property defaults, normalization, and validation via Builder

bryanstern:
This is a first pass at implementing a solution for #132 using the proposed AutoValue builder properties google/auto#501.

If an AutoValue class specifies a builder with the `@AutoValue.Builder` annotation, the generated `TypeAdapter#read` implementation will make use of the builder. If the builder performs any [normalization](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#validate) or [validation](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-normalize-modify-a-property-value-at-build-time), it will run.

The TypeAdapter default setters have been removed as builders allow the user to specify defaults using the [AutoValue convention](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#default). Annotate the `builder()` method with `@GsonDefaultBuilder` to specify which method AutoValueGson should use to get an instance of the builder.

If your builder specifies multiple `build()` or unconventional build methods (not named `build`), you indicate which method in the Builder should be used with `@GsonBuild`.


README.md
```diff
@@ -71,6 +68,67 @@ see [Gson Changelog](https://github.com/google/gson/blob/master/CHANGELOG.md#ver
...
+and declaring defaults there.
+
+```java
+@AutoValue public abstract class Foo {
+  abstract Optional<String> bar();
+  abstract Optional<String> baz();
+  abstract String quux();
+  
+  @GsonDefaultBuilder
+  public static builder() {
```
hzsweers: nit: missing the return type on the method signature

auto-value-gson/build.gradle
```diff
@@ -9,7 +9,7 @@ targetCompatibility = JavaVersion.VERSION_1_8
 dependencies {
     api 'com.google.code.gson:gson:2.8.1'
     implementation 'com.squareup:javapoet:1.9.0'
-    api 'com.google.auto.value:auto-value:1.4.1'
+    api 'com.github.bryanstern.auto:auto-value:extension-builder-support-SNAPSHOT'
```
hzsweers: nit: maybe add a comment marking this as temporary

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
```diff
@@ -285,5 +286,15 @@ public LimitedContext(ProcessingEnvironment processingEnvironment, TypeElement a
     @Override public Set<ExecutableElement> abstractMethods() {
       return null;
     }
+
+    @Override
+    public boolean hasBuilder() {
+      return false;
+    }
+
+    @Override
```
hzsweers: nit: `@Nullable`

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/GsonBuild.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * If present, will indicate which builder build method to use. Only necessary if the build method
+ * is not named "build".
```
hzsweers: nit: `{@code build}`

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/GsonDefaultBuilder.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * If present, indicates that the annotated method should be used for retrieving an instance of the
+ * AutoValue.Builder. This provides a mechanism for supplying default values.
+ */
+@Retention(SOURCE)
+@Target(METHOD)
+public @interface GsonDefaultBuilder {
```
hzsweers: Wondering if this should maybe be called something more like `AutoValueGsonBuilder` to make it specific

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -97,6 +99,10 @@ public Property(String humanName, ExecutableElement element) {
       annotations = buildAnnotations(element);
 
       typeAdapter = getAnnotationValue(element, GsonTypeAdapter.class);
+      TypeMirror returnType = element.getReturnType();
+      instanceOfOptional = !TypeKind.TYPEVAR.equals(returnType.getKind())
```
hzsweers: This is a bit limited since it only supports guava and java 8's. Could we make it more generic? (Unless AutoValue only supports these two, then it's fine. My memory's fuzzy)

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -304,25 +286,6 @@ public String generateClass(Context context, String className, String classToExt
     return fields.build();
   }
 
-  private Map<Property, FieldSpec> createDefaultValueFields(List<Property> properties) {
```
hzsweers: 😍

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+    MethodSpec.Builder readMethod = MethodSpec.methodBuilder("read")
+        .addAnnotation(Override.class)
+        .addModifiers(PUBLIC)
+        .returns(autoValueClassName)
+        .addParameter(jsonReader)
+        .addException(IOException.class);
+
+    ClassName token = ClassName.get(JsonToken.NULL.getDeclaringClass());
+
+    readMethod.beginControlFlow("if ($N.peek() == $T.NULL)", jsonReader, token);
```
hzsweers: side note - always annoyed me that we do this vs just having people call `.nullSafe()` on the adapters if they want them to be so

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+    readMethod.addStatement("$N.nextNull()", jsonReader);
+    readMethod.addStatement("return null");
+    readMethod.endControlFlow();
+
+    Optional<ExecutableElement> defaultBuilder = findDefaultBuilder(autoValueClassElement, builderContext.builderClass());
+
+    ClassName builderClassName = ClassName.get(builderContext.builderClass());
+    FieldSpec builder = FieldSpec.builder(builderClassName, "_builder").build();
```
hzsweers: Nit: this project doesn't use the `_` prefix convention, let's not use it here and in the other places below

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+      readMethod.beginControlFlow("case $S:", prop.serializedName());
+      Set<ExecutableElement> setters = builderContext.setters().get(prop.humanName);
+
+      // search overloaded methods for a type match
+      Optional<CodeBlock> setterBlock = setters.stream()
+          .filter(e -> ClassName.get(e.getParameters().get(0).asType()).equals(prop.type))
+          .findFirst()
+          .map(method ->
+              CodeBlock.builder()
+                  .addStatement("_builder.$N($N.read($N))", method.getSimpleName(), adapters.get(prop), jsonReader)
```
hzsweers: You can keep this safe by replacing the builder ref in the statement with `$N` and passing the `builder` parameterspec as the argument for it. JavaPoet will do the matching under the hood

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+                  .build());
+
+      // if the property is Optional<T>, see if there is an overloaded setter that accepts T
+      if (!setterBlock.isPresent() && prop.instanceOfOptional) {
+        final ClassName optionalPropTypeArgument = (ClassName) ((ParameterizedTypeName) prop.type).typeArguments.get(0);
+        setterBlock = setters.stream()
+            .filter(e -> ClassName.get(e.getParameters().get(0).asType()).equals(optionalPropTypeArgument))
+            .findFirst()
+            .map(method ->
+            {
```
hzsweers: nit - code style

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+              String propertyVarName = prop.humanName + "Property";
+              return CodeBlock.builder()
+                  .addStatement("$T $L = $N.read($N)", prop.type, propertyVarName,
+                      adapters.get(prop), jsonReader)
+                  .beginControlFlow("if ($L.isPresent())", propertyVarName)
+                  .addStatement("_builder.$N($L.get())", method.getSimpleName(), propertyVarName)
```
hzsweers: same here with using the `builder` param and making the reference in the statement use `$N`

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+      readMethod.addStatement("break");
+      readMethod.endControlFlow();
+    }
+
+    // skip value if field is not serialized...
+    readMethod.beginControlFlow("default:");
+    readMethod.addStatement("$N.skipValue()", jsonReader);
+    readMethod.endControlFlow();
+
+    readMethod.endControlFlow(); // switch
```
hzsweers: good use of comment! Hadn't thought of it before but it definitely helps a lot with readability

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -704,4 +750,101 @@ private static void buildParameterizedTypeArguments(CodeBlock.Builder block, Typ
       block.add("$T.class", typeArg);
     }
   }
+
+  private static Optional<ExecutableElement> findDefaultBuilder(TypeElement autoValueClass, TypeElement builderClass) {
```
hzsweers: Nit: I prefer using the type directly + `@Nullable` rather than Optional when it's just implementation details and not differentiating between `null` and `absent` (that is, the sort of "three" implications of Optionals). Same with the other find method below.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -704,4 +750,101 @@ private static void buildParameterizedTypeArguments(CodeBlock.Builder block, Typ
...
+    } else if (annotatedMethods.size() == 1) {
+      return Optional.of(Iterables.getOnlyElement(annotatedMethods));
+    }
+
+    return methods.stream()
+        .filter(e -> e.getSimpleName().contentEquals("build"))
+        .findFirst();
+  }
+
+  static boolean validateTypeAdapter(Context context) {
```
hzsweers: since this returns a boolean, I think it should be named with the `is` prefix. The verb tense makes me think it throws if it's not, but maybe that's just me

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
rharter: agreed

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+    readMethod.addStatement("$N.nextNull()", jsonReader);
+    readMethod.addStatement("return null");
+    readMethod.endControlFlow();
+
+    Optional<ExecutableElement> defaultBuilder = findDefaultBuilder(autoValueClassElement, builderContext.builderClass());
+
+    ClassName builderClassName = ClassName.get(builderContext.builderClass());
+    FieldSpec builder = FieldSpec.builder(builderClassName, "_builder").build();
+    if (defaultBuilder.isPresent()) {
+      ClassName baseClass = ClassName.get(autoValueClassElement);
```
rharter: We already pass in the `autoValueClassName` as a `TypeName`, so if the `autoValueClassElement` is only used to get the name here, lets remove the parameter.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+
+      for (String alternate : prop.serializedNameAlternate()) {
+        readMethod.addCode("case $S:\n", alternate);
+      }
+      readMethod.beginControlFlow("case $S:", prop.serializedName());
+      Set<ExecutableElement> setters = builderContext.setters().get(prop.humanName);
+
+      // search overloaded methods for a type match
+      Optional<CodeBlock> setterBlock = setters.stream()
+          .filter(e -> ClassName.get(e.getParameters().get(0).asType()).equals(prop.type))
```
rharter: Just looking on my phone, but is this right?  Don't you also have to check the name for a match?  If I have an object with two String properties, won't this always find the first one?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -587,6 +519,120 @@ public MethodSpec createReadMethod(ClassName className,
...
+                  .addStatement("$T $L = $N.read($N)", prop.type, propertyVarName,
+                      adapters.get(prop), jsonReader)
+                  .beginControlFlow("if ($L.isPresent())", propertyVarName)
+                  .addStatement("_builder.$N($L.get())", method.getSimpleName(), propertyVarName)
+                  .endControlFlow()
+                  .build();
+            });
+      }
+
+      readMethod.addCode(setterBlock.get()); // S
```
rharter: What does this comment mean?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -704,4 +750,101 @@ private static void buildParameterizedTypeArguments(CodeBlock.Builder block, Typ
...
+        .filter(e -> e.getParameters().isEmpty())
+        .findFirst();
+  }
+
+  /**
+   * Searches the builder for an method has no parameters, returns the autovalue class, is not
+   * private, and is not static. If there is a valid method annotated with {@link GsonBuild}, return
+   * that method. Otherwise, return the method named "build" if present.
+   */
+  private static Optional<ExecutableElement> findBuildMethod(
```
rharter: This should follow the same method that AutoValue uses to identify build methods in a builder.  That is, a non-private, non-static method with no args that returns the annotated type.  The method need not be named `build` (`create` is another common option).  If there is more than one (both `create` and `build` and `createWithDefaults` and `createByIncrementingCount`, etc) then you can use the annotation to decide which one.

rharter: This may seem nit-picky, but AutoValue sort of prides itself on not caring what you name your methods.  We shouldn't get in the way of that.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
rharter: agreed.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/GsonBuild.java
rharter: method names don't matter, use appropriate identification techniques mentioned above.

build.gradle
```diff
@@ -17,6 +17,9 @@ allprojects {
         maven {
             url 'http://oss.sonatype.org/content/repositories/snapshots'
         }
+        maven {
+            url 'https://jitpack.io'
```
rharter: please TODO comment that this needs to be removed when real support is released.

example/src/main/java/com/ryanharter/auto/value/gson/example/Person.java
```diff
@@ -19,8 +20,11 @@
 
     public abstract Address address();
 
+    @GsonDefaultBuilder
```
rharter: The need for this will go away when builder methods are properly identified.

example/src/main/java/com/ryanharter/auto/value/gson/example/Person.java
```diff
@@ -39,6 +43,14 @@ public static Builder builder() {
 
         public abstract Builder address(Address address);
 
-        public abstract Person build();
+        abstract Person autoBuild();
+
+        public Person build() {
```
rharter: Here's an example where the annotation would be needed.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: Good call. I forgot to cover the all the Optional variants in java.util

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: Makes sense. I saw the original usage to avoid variable name conflicts and mistakenly thought it applied here too.

rharter: You can use Java Poet's NameAllocator to take care of that.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -172,67 +178,43 @@ public boolean nullable() {
...
-        }
-      }
-    }
+    // validate @GsonDefaultBuilder usage
+    Set<ExecutableElement> annotatedElements = ElementFilter
+        .methodsIn(context.autoValueClass().getEnclosedElements())
+        .stream()
+        .filter(MoreElements.hasModifiers(STATIC)::apply)
+        .filter(e -> !MoreElements.hasModifiers(PRIVATE).apply(e))
+        .filter(e -> MoreElements.isAnnotationPresent(e, GsonDefaultBuilder.class))
```
bryanstern: Makes sense

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: Just noise from notes I was making while working. Will remove.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: It should be ok because `builderContext.setters()` is a multimap of properties to their setter(s).

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: We need it to pick out the build method in the builder. See the call to `findBuildMethod` near the end of this method.

bryanstern:
Incorporated your feedback. I totally agree we should infer the static `builder()` methods where we can and be less picky about `build` vs `create`.

liujoshua:
Any updates on this? I'm using @AutoValue.Builder and providing default/initial values on the builder, e.g. 
empty Guava Immutable collections.

bryanstern:
@liujoshua still blocked on the AutoValue project. https://github.com/google/auto/pull/501

bryanstern:
Closing this in favor of #224

PR #141, opened 2017-08-07
Update README to include mutable adapter flag.

alexfu:
Adds missing documentation on how to enable mutable TypeAdapters.


hzsweers:
Thanks!

PR #144, opened 2017-08-18
Generate Nullable annotation at constructor arguments.

mik9:
Generate @Nullable annotation at constructor arguments.

AutoValue itself does this for generated class.
Also when we use @ParametersAreNonnullByDefault in project then IDE and Spotbugs will complain when we try to pass null as constructor argument which not annotated with @Nullable (but appropriate abstract method does annotated).


hzsweers:
Could you add a description?

mik9:
Updated description.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -157,15 +157,23 @@ public boolean shouldDeserialize() {
...
+      return !nullableAnnotation().isEmpty();
+    }
+
+    public String nullableAnnotation() {
+      for (String annotationString : annotations) {
+        if (annotationString.equals("@Nullable") || annotationString.endsWith(".Nullable")) {
+          return annotationString;
+        }
+      }
+      return "";
```
hzsweers: This should return null if not found, not an empty string

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -157,15 +157,23 @@ public boolean shouldDeserialize() {
     }
 
     public boolean nullable() {
-      return annotations.contains("Nullable");
+      return !nullableAnnotation().isEmpty();
```
hzsweers: Can we pre-calculate this and save it?

mik9: Sure, I just took code from AutoValue's Property class.

mik9:
Hello, any more changes I need to make to allow this to be merged?

hzsweers:
Sorry I didn't see you'd updated it

hzsweers:
Thanks!

mik9:
Np, thanks! And same PR for AutoValue Gson: https://github.com/rharter/auto-value-parcel/pull/124

Issue #123, opened 2017-03-21
Add support for Optional


panfngsugar:
AutoValue now supports Optional properties. Can we also add support for Optional Properties?

acdebaca:
Would a pull request be considered for this feature if I provided one?

The current issue is the null guard in the generated code. Input:
```java
@AutoValue
abstract class Offer {
	abstract String plainString();
	abstract Optional<String> optionalString();

	static TypeAdapter<Offer> typeAdapter(final Gson gson) {
		return new AutoValue_Offer.GsonTypeAdapter(gson);
	}
}
```

Output, `read` method only:
```java
    @Override
    public Offer read(JsonReader jsonReader) throws IOException {
      if (jsonReader.peek() == JsonToken.NULL) {
        jsonReader.nextNull();
        return null;
      }
      jsonReader.beginObject();
      String plainString = null;
      Optional<String> optionalString = null; // This would be Optional.empty()
      while (jsonReader.hasNext()) {
        String _name = jsonReader.nextName();
        if (jsonReader.peek() == JsonToken.NULL) { // Null would propagate to optional adapter
          jsonReader.nextNull();
          continue;
        }
        switch (_name) {
          case "plainString": {
            plainString = plainStringAdapter.read(jsonReader);
            break;
          }
          case "optionalString": {
            optionalString = optionalStringAdapter.read(jsonReader);
            break;
          }
          default: {
            jsonReader.skipValue();
          }
        }
      }
      jsonReader.endObject();
      return new AutoValue_Offer(plainString, optionalString); // optionalString = null
    }
```
As a result, `optionalString` is never processed, and the parent `$AutoValue_Offer` constructor throws a validation exception because the argument is null.

Note: I wrote my own `TypeAdapterFactory` implementation that works with `java.util.Optional`. It would be trivial to add another for Guava `Optional` as well.

rharter:
Yeah, a PR would be appreciated for this.

acdebaca:
@rharter Happy to do it. Before proceeding, I'd like your guidance on one issue. auto-value-gson would have to include a `TypeAdapter` (and/or corresponding factory) for Java and Guava `Optional`.

1. Are you comfortable with furnishing `TypeAdapter` implementations as part of the library? These classes need to be present at runtime. I don't think that there is precedent for that in this project. If so, I assume they would go in the `com.ryanharter.auto.value.gson` package, but in which artifact?

[...]

PR #148, opened 2017-09-30
Support optional

acdebaca:
This is a first-cut implementation for issue #123, adding support for optional types. This implementation supports the following optional types:

`com.google.common.base.Optional`
`java.util.Optional`
`java.util.OptionalDouble`
`java.util.OptionalInt`
`java.util.OptionalLong`

Regression tests have been added to cover various combinations of features.


auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -1366,13 +1366,22 @@
...
 
+    // Collections default to empty
     assertAbout(javaSources())
         .that(Arrays.asList(nullable, source))
         .withCompilerOptions("-A" + COLLECTIONS_DEFAULT_TO_EMPTY + "=true")
         .processedWith(new AutoValueProcessor())
         .compilesWithoutError()
         .and()
         .generatesSources(expected);
+
```
acdebaca: All properties in this test, including collection types, are wrapped with an `Optional`. Because the current implementation defaults optional values to the corresponding empty/absent value, no distinction is made between `Optional`-wrapped collection types and any other `Optional`-wrapped type. To confirm that this is the case, this test is run both with and without defaulting collection-typed properties to empty.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -527,16 +537,29 @@ public TypeSpec createTypeAdapter(Context context, ClassName className, ClassNam
   public List<MethodSpec> createDefaultMethods(ClassName gsonTypeAdapterName, List<Property> properties) {
     List<MethodSpec> methodSpecs = new ArrayList<>(properties.size());
     for (Property prop : properties) {
-      ParameterSpec valueParam = ParameterSpec.builder(prop.type, "default" + upperCamelizeHumanName(prop)).build();
```
acdebaca: When using AutoValue with builders and optional properties, the builder setters can take the wrapped type. In that vein, the setters are generated accepting the wrapped type. Note that an `Optional<Integer>` property would yield a setter that takes an `Integer` boxed type, whereas `OptionalInt` would yield a setter that takes a `int` primitive.

The generated setter code allows setters to accept null by using the `ofNullable` static method before assigning to the type adapter field.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -688,6 +711,10 @@ private CodeBlock getDefaultValue(Property prop, FieldSpec field) {
         return CodeBlock.of("$T.valueOf(null)", field.type);
       }
     }
```
acdebaca: This change asserts that being an optional type takes precedence over being nullable when determining what default value to assign a property. As a result, a nullable optional property will always default to the appropriate empty/absent value during deserialization when the entry is missing from the JSON object or if it is present with a `null` value.

I believe that this is sensible but note that it is somewhat inconsistent with AutoValue. A nullable optional property in AutoValue will default the property to null if the corresponding builder setter is never called.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -34,8 +33,7 @@ public void setup() {
         + "}");
   }
```
acdebaca: For more comprehensive testing, I took the optional property tests out of the `simple()` method and instead created entirely new tests based on `simple()` in which all properties are optional. Java and Guava optionals are tested in separate methods (with Java including the primitive optional wrappers), and each of Java/Guava have separate tests for w/ and w/o default setter cases.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -481,26 +493,24 @@ public TypeSpec createTypeAdapter(Context context, ClassName className, ClassNam
         continue;
       }
```
acdebaca: Here, we substitute the property type for its type parameter. All downstream processing thereby stays the same for handling parameterized types, primitives, etc.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -83,12 +84,60 @@
           .addMember("comments", "$S", GENERATED_COMMENTS)
           .build();
 
-  private static final ImmutableSet<TypeName> OPTIONAL_RAW_TYPENAMES = ImmutableSet.of(
-      TypeName.get(com.google.common.base.Optional.class),
-      TypeName.get(java.util.Optional.class),
-      TypeName.get(java.util.OptionalInt.class),
-      TypeName.get(java.util.OptionalDouble.class),
-      TypeName.get(java.util.OptionalLong.class));
```
acdebaca: In the first commit, there was a lot of duplication of processing logic to determine how to treat an optional property. This class consolidates that logic in the static factory method `fromPropertyType`.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -270,15 +270,15 @@ public void simple() {
         + "      jsonWriter.name(\"o\");\n"
         + "      oAdapter.write(jsonWriter, object.o());\n"
         + "      jsonWriter.name(\"p\");\n"
```
acdebaca: Empty/absent properties are serialized as null, or the default value for wrapped primitives.
User-created objects will be empty/absent in the following cases:
1. `Optional.empty()` is passed to the static factory method (when not using builders).
2. When using builders, the setter is not called before building and default setters are not enabled on the type adapter.
3. When using builders, the setter is not called before building and default setters are enabled but the default setter is never invoked.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -1169,18 +1360,867 @@ public void simple() {
...
+        + "            break;\n"
+        + "          }\n"
+        + "          default: {\n"
+        + "            jsonReader.skipValue();\n"
+        + "          }\n"
+        + "        }\n"
+        + "      }\n"
+        + "      jsonReader.endObject();\n"
+        + "      return new AutoValue_Test(a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q);\n"
+        + "    }\n"
```
acdebaca: Note that default setters take the wrapped type. Because null can be passed, `ofNullable` is used here.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -1169,18 +1360,867 @@ public void simple() {
...
+        + "    public GsonTypeAdapter setDefaultA(String defaultA) {\n"
+        + "      this.defaultA = Optional.ofNullable(defaultA);\n"
+        + "      return this;\n"
+        + "    }\n"
+        + "    public GsonTypeAdapter setDefaultB(int[] defaultB) {\n"
+        + "      this.defaultB = Optional.ofNullable(defaultB);\n"
+        + "      return this;\n"
+        + "    }\n"
```
acdebaca: Wrapped boxed type `Optional<Integer>`, so `setDefaultC` takes an `Integer`

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -1169,18 +1360,867 @@ public void simple() {
...
+        + "    public GsonTypeAdapter setDefaultM(String defaultM) {\n"
+        + "      this.defaultM = Optional.ofNullable(defaultM);\n"
+        + "      return this;\n"
+        + "    }\n"
+        + "    public GsonTypeAdapter setDefaultN(List<String> defaultN) {\n"
+        + "      this.defaultN = Optional.ofNullable(defaultN);\n"
+        + "      return this;\n"
+        + "    }\n"
```
acdebaca: Wrapped primitives, so default setters take the primitive.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
acdebaca: Also note that this defaults `Optional` of a collection type to `Optional.empty()`/`Optional.absent()` and not to an empty collection, regardless of the `defaultCollectionsToEmpty` compiler option. For example, a property of type `Optional<ImmutableSet<String>>`) defaults to `Optional.empty()` and not `Optional.of(ImmutableSet.of())`.

rharter: This is the change I'm a bit concerned about, and would like to hear the reasoning for the inconsistency with AutoValue.  My understanding is that this basically negates the `@Nullable` annotation completely, as the value could never actually be null.  In practice, this might not matter, as I'm not sure why one would have a nullable optional field, as they seem mutually exclusive.

If the decision is arbitrary, however, I lean towards consistency with AutoValue (if it doesn't complicate things immensely).  I'd appreciate some other input here.

Re: the optional collections, I think that's probably the wrong approach, but, again, would be open to discussion.  My thinking is that `Optional` basically takes the place of `null` to provide better null safety, and the point of the `defaultCollectionToEmpty` option was to provide empty collections in cases where the JSON field was missing.

Most other places users can simply swap `@Nullable` for `Optional` and functionality stays the same, but this changes functionality, basically making the `defaultCollectionToEmpty` a useless compiler option, since it sounds like it does nothing for Optional field.  

Since that compiler option was basically providing a way for users to bandage broken APIs (we always say that the correct fix is to fix your server if it's returning null when you don't want it), we should make sure the bandage still works if users upgrade to `Optional`.

rharter:
This looks really good, overall, just a couple of discussion points.

gabrielittner:
Might be worth it to bump https://github.com/google/auto/issues/359 so that AutoValue and this can share the same logic around this.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
acdebaca: (please disregard my deleted comment - I was brainstorming and inadvertently submitted incoherent gibberish).

For the `@Nullable Optional` case:
To be clear, the inconsistency is with `@AutoValue.Builder` and its behavior when a property is not set. I concur that, for consistency with AV and for uniformity in the treatment of `@Nullable`, the default here should be null.

acdebaca: For `Optional` collections, I still think that `empty()` is the correct deserialization default. It actually won't change the functionality when swapping `@Nullable` for `Optional` for a collection-valued field, unless I am misreading the code.

`@Nullable` currently takes precedence over `defaultCollectionsToEmpty`. As a result, av-gson currently provides two different, reasonable ways to handle null for collection-valued properties. When `@Nullable` is present on such a property, `defaultCollectionsToEmpty` is not considered for that property.

As with my previous points about `@Nullable Optional`, this use of `@Nullable` is uniform without regard to the property type. If `@Nullable` is omitted, the user can opt-in for an empty collection via the compiler argument. This behavior is also consistent with `@AutoValue.Builder`, whereby omitting a `@Nullable` property results in null by default, regardless of the corresponding property type.

Moreover, when a user wraps a collection-typed property with `Optional`, they are changing the type of property. Unlike `@Nullable`, which is just an annotation, a type change alters the semantics and syntax of how the value must be handled, as Java doesn't permit treating an `Optional` as a proxy for its wrapped type. Because the indirection is explicit, I can't imagine that a user would expect that whatever is wrapped gets special treatment, when the outermost class has a defined way to represent nullness.

Said another way: If the user wants an empty collection by default, why wrap it with `Optional` at all? I suppose there could be other reasons for making a field `Optional` which are unrelated to gson deserialization.

I'm happy to make the changes either way. However, if you do decide that `Optional<Collection<T>>` should default to empty collection, I suggest also changing the behavior of `@Nullable Collection<T>` so that they are consistent.

acdebaca: Hi @rharter - just pinging you to get your feedback on our discussion above. I'd be happy to implement the changes as you see fit. Above was just my 2c.

acdebaca:
Hi @rharter - I know it's been a few months, but I'd still love to get this code up-streamed if possible. Please review and let me know how you'd like for me to proceed. Thanks!

rharter:
Hi @acdebaca, sorry for letting this languish.  I think you've cleared things up for me and I'm happy to merge this once the conflicts are resolved.  Thanks!

acdebaca:
@rharter Great! I'll update this discussion when the changes are ready.

epool:
@acdebaca firstly thank you so much for this helpful PR, but, do you have an ETA for the conflicts resolution? Thanks again!

acdebaca:
@epool I don't have an ETA on this, but I'll see if I can get them fixed in one sitting today or tomorrow.

kaibolay:
@acdebaca I'm looking for this feature. I hope you find some time to clean up the merge conflicts. 

If you don't have time, would you mind if someone else (e.g. me) tried to resolve the conflicts and get this merged?

epool:
@kaibolay I ended up using kotlin data classes instead of autovalue.

rharter:
@kaibolay feel free to check out this branch and fix the merge conflicts and resubmit. Would be appreciated!

hzsweers:
I think we should make a call for now to just not support these. Modeling absence in general Java code is one thing, but modeling it in JSON is very different, and I'd rather opt for no solution rather than a limited one and encourage better support for normalization from AutoValue via #132.

PR #150, opened 2017-10-11
Update provided dependency in readme

bejibx:


rharter:
@JakeWharton I was just updating that in #151

PR #156, opened 2017-11-23
Generated code optimizations

hzsweers:
This makes two optimizations to generated code:
- Don't write keys for optional values that are null (otherwise the keys are just taking up space on the wire)
- Reuse adapter instances for the same types rather than always generating one adapter per type.

Also updated a few dependencies opportunistically along the way.

av-moshi already does the optional check. Once things are reviewed/done here, I'll open a PR there for the reused adapter fields logic


auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -143,62 +142,33 @@ public void simple() {
...
-        + "    private final TypeAdapter<String> iAdapter;\n"
-        + "    private final TypeAdapter<List<String>> jAdapter;\n"
-        + "    private final TypeAdapter<String> kAdapter;\n"
-        + "    private final TypeAdapter<List<String>> lAdapter;\n"
-        + "    private final TypeAdapter<String> mAdapter;\n"
-        + "    private final TypeAdapter<List<String>> nAdapter;\n"
-        + "    private final TypeAdapter<Map<String, Map<String, Map<String, Map<String, "
-        + "Map<String, String>>>>>> oAdapter;\n"
+        + "    private final TypeAdapter<String> java_lang_StringAdapter;\n"
+        + "    private final TypeAdapter<int[]> int__Adapter;\n"
```
hzsweers: Some of these names don't quite come out perfectly readible, but I think it's sort of ok since it's generated code and the TypeNames themselves still hash correctly

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -143,62 +142,33 @@ public void simple() {
...
+        + "    private final TypeAdapter<Integer> int_Adapter;\n"
+        + "    private final TypeAdapter<ImmutableMap<String, Number>> com_google_common_collect_ImmutableMap_java_lang_String__java_lang_Number_Adapter;\n"
+        + "    private final TypeAdapter<Set<String>> java_util_Set_java_lang_String_Adapter;\n"
+        + "    private final TypeAdapter<Map<String, Set<String>>> java_util_Map_java_lang_String__java_util_Set_java_lang_String__Adapter;\n"
+        + "    private final TypeAdapter<List<String>> java_util_List_java_lang_String_Adapter;\n"
+        + "    private final TypeAdapter<Map<String, Map<String, Map<String, Map<String, Map<String, String>>>>>> java_util_Map_java_lang_String__java_util_Map_java_lang_String__java_util_Map_java_lang_String__java_util_Map_java_lang_String__java_util_Map_java_lang_String__java_lang_String_____Adapter;\n"
```
hzsweers: ones like this make me wonder if we should shorten package names to first letters or something, like dagger does

rharter: I would say just drop the package name all together and use the simple name.  Most of the time that will be fine, and the `NameAllocator` will take care of collisions (so you'll have `StringAdapter` and `StringAdapter1`). And since you're storing them in a map by type, so the generator will be able to resolve the correct adapter.

JakeWharton:
Keys for null values are deferred and only written to the wire when serializeNulls is true. Can we emulate that instead?

rharter:
By moshi, you mean?  When setting a value to null?

JakeWharton:
No, Gson does this. I don't know of Moshi has a control for this or not.

rharter:
Ahh, yes, sorry, wrong repo :)  So that would allow us to ignore it entirely, and probably explains why we have this in AVM but not AVG.

hzsweers:
TIL. Will remove

hzsweers:
Updated with the removed null checks in 8abe120 and shortened typenames in 427319a. I didn't make it toooo dumb with the naming, but should cover most cases. 

Oddly, `generatesCorrectDefaultCharPrimitiveValue` always fails with really strange output. I'm wondering if it's a bug in compile-testing.

```
Diffs:
======

Found 1 nodes that differed in expected and actual trees. 

> Difference in expected tree and actual tree.
  Expected node: Line 46 COMPILATION_UNIT->CLASS(AutoValue_Test)->CLASS(GsonTypeAdapter)->METHOD(read)->BLOCK(non-static)->VARIABLE(c)->CHAR_LITERAL(?)
  Actual node: Line 46 COMPILATION_UNIT->CLASS(AutoValue_Test)->CLASS(GsonTypeAdapter)->METHOD(read)->BLOCK(non-static)->VARIABLE(c)->CHAR_LITERAL(?)
  Expected literal value to be <?> but was <?>.
```

That line is `char c = '?';`, which is indeed identical in the expected and actual outputs. Going to ping someone on the core libraries team and try a different character

hzsweers:
Any thoughts on the `generatesCorrectDefaultCharPrimitiveValue` issue?

rharter:
As long as they're the same unicode value then that definitely looks like a bug in `compile-testing`.  Did you hear anything back from Greg?

hzsweers:
Pinged someone on the core libraries team

ronshapiro:
My first thought is that maybe there's something weird about how `javac` stores `JCLiteral.getValue()` [here](https://github.com/google/compile-testing/blob/49c254f02a0d9c9c9ebbedd7c7cd9635ecd699d5/src/main/java/com/google/testing/compile/TreeDiffer.java#L645). I'll try to see if I can repro

ronshapiro:
Ah, what I thought was too obvious appears to be the case: My guess is that when you were updating the tests in `21d9fcb1` you just copied the content directly from your terminal, and the terminal printed a `?` when it didn't have any other glyph for the null character `\0` (which is what is at HEAD). 

When I run the tests, I get the following error:

```
Expected node: Line 46 COMPILATION_UNIT->CLASS(AutoValue_Test)->CLASS(GsonTypeAdapter)->METHOD(read)->BLOCK(non-static)->VARIABLE(c)->CHAR_LITERAL(?)
Actual node: Line 46 COMPILATION_UNIT->CLASS(AutoValue_Test)->CLASS(GsonTypeAdapter)->METHOD(read)->BLOCK(non-static)->VARIABLE(c)->CHAR_LITERAL()
Expected literal value to be <?> but was <>.
```

Which seems expected. When I revert that line so it's back to `char c = '\0';` the test passes again.

rharter:
hah, that makes sense.  Thanks @ronshapiro!

hzsweers:
I should've checked master :). It handled it already, fixed in 542f3eb

PR #160, opened 2017-12-07
Implement @GenerateTypeAdapter support

hzsweers:
This implemented support for `@GenerateTypeAdapter`, which is an annotation that can be used to indicate that the developer wants the generated typeadapater to be generated as a top-level class in the same package. This also includes a runtime reflective typeadapterfactory (loosely based on Jake's AutoGson + Butter Knife's lookup logic).

What this means is that you can now trigger generation with this annotation OR a static TypeAdapter-returning method.

You can use both a static method and the annotation, but if you do, the static method implementation will need to point to the public top-level class.

If the annotation is present, no intermediary AutoValue class will be generated either.

The adapter lookups are cached similar to how Butter Knife's bindings are cached.

The runtime factory is intended to be an alternative to generating a factory at the cost of a little (aforementioned cached) reflection. The benefit is that if you have a huge multi-module repo, you don't have to manually manage wiring up generated factories to some top level gson instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,80 @@ android {
 }
 ```
 
+## @GenerateTypeAdapter
+
+There is an annotation in the `auto-value-gson-annotations` artifact called `@GenerateTypeAdapter`. This annotation
+can be set on types to indicate to the extension that you want the generated adapter to be a top level class in the same
+package. The name of this class will be the AutoValue class's name plus `_GsonTypeAdapter` suffix.
+
+Types annotated with this can also be (de)serialized dynamically at runtime with a provided runtime `TypeAdapterFactory`
+implementation in the annotation called `FACTORY`. The type name and generated typeadapter class's name *must not be obfuscated*
+for this to work.
+
+When this annotation is used, there will be no intermediate AutoValue class generated (as opposed to the default logic, 
+which generates an intermediate class and generates the `TypeAdapter` as a static inner class of it).
+
+`@GenerateTypeAdapter` is compatible with the factory approach above, just make your static method's implementation
+point to it. It can also be an alternative to it if you use the runtime factory, particularly if you 
+have a multimodule project and are willing to accept a small amount of (heavily cached) reflection.
+
+The generated class will have the same parameters as if it were the inner class. If it's generic, its constructor
+accepts a `Gson` instance and `TypeToken` of the generics. If it's not generic, it's just a `Gson` instance.
+
+Example usage:
+
+```java
+@GenerateTypeAdapter
+@AutoValue
+public class Foo {
+  // ...
+  public static TypeAdapter<Foo> typeAdapter(Gson gson) {
+    return new Foo_GsonTypeAdapter(gson);
+  }
+}
+
+// Generates
+public final class Foo_GsonTypeAdapter extends TypeAdapter<Foo> {
+  public Foo_GsonTypeAdapter(Gson gson) {
+    //...
+  }
+}
+
+// Or with generics
+@GenerateTypeAdapter
+@AutoValue
+public class Foo<T> {
+  // ...
+  public static TypeAdapter<Foo> typeAdapter(Gson gson, TypeToken<? extends Foo<T>> typeToken) {
+    return new Foo_GsonTypeAdapter(gson, typeToken);
+  }
+}
+
+// Generates
+public final class Foo_GsonTypeAdapter extends TypeAdapter<Foo> {
+  public Foo_GsonTypeAdapter(Gson gson, TypeToken<? extends Foo<T>> typeToken) {
+    //...
+  }
+}
+
+// Using the runtime FACTORY
+new GsonBuilder()
+    .addTypeAdapterFactory(GenerateTypeAdapter.FACTORY)
+    .build()
+    .toJson(myFooInstance);
+```
+
+Proguard rules for obfuscation:
+
+```
+# Retain generated classes that end in the suffix
+-keepnames class **_GsonTypeAdapter
+
+# Prevent obfuscation of types which use @GenerateTypeAdapter since the simple name
+# is used to reflectively look up the generated adapter.
+-keepnames @com.ryanharter.auto.value.gson.GenerateTypeAdapter class *
+```
+
 ## Download
 
 Add a Gradle dependency to the `apt` and `provided` configuration.
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ Add a Gradle dependency to the `apt` and `provided` configuration.
 
 ```groovy
 annotationProcessor 'com.ryanharter.auto.value:auto-value-gson:0.6.0'
-provided 'com.ryanharter.auto.value:auto-value-gson-annotations:0.6.0'
+compile 'com.ryanharter.auto.value:auto-value-gson-annotations:0.6.0'
 ```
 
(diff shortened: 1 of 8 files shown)


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
```diff
@@ -135,15 +137,24 @@ private TypeSpec createTypeAdapterFactory(
         .returns(result)
         .addStatement("Class<$T> rawType = (Class<$T>) $N.getRawType()", t, t, type);
 
-    for (int i = 0, elementsSize = elements.size(); i < elementsSize; i++) {
-      Element element = elements.get(i);
+    List<Map.Entry<Element, Optional<ExecutableElement>>> properties = elements.stream()
+        .collect(Collectors.toMap(e -> e, this::getTypeAdapterMethod))
+        .entrySet()
+        .stream()
```
JakeWharton: collect to map, entry set, and stream all can be replaced by mapping to a pair / entry set directly.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
```diff
@@ -135,15 +137,24 @@ private TypeSpec createTypeAdapterFactory(
...
+        .filter(entry -> entry.getValue().isPresent())
```
JakeWharton: Java 9 destroys this terrible pattern with `flatMap(Optional::stream)`. I'm tempted to even say that you should replace it with `flatMap(entry -> entry.map(Stream::of).orElse(Stream.empty())` to avoid using `isPresent` and `get`.

example/src/main/java/com/ryanharter/auto/value/gson/example/WebResponse.java
```diff
@@ -5,10 +5,12 @@
 import com.google.gson.TypeAdapter;
 import com.google.gson.reflect.TypeToken;
 
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
```
JakeWharton: Unused

example/src/main/java/com/ryanharter/auto/value/gson/example/WebResponse.java
```diff
@@ -5,10 +5,12 @@
...
 import java.util.List;
 import java.util.Map;
 
-@AutoValue public abstract class WebResponse<T> {
+@AutoValue
+public abstract class WebResponse<T> {
```
JakeWharton: Revert

auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+ * TypeAdapter in a separate class rather than an inner class of an intermediate AutoValue-generated
+ * class hierarchy.
+ */
+@Inherited
+@Retention(RUNTIME)
+@Target(TYPE)
+public @interface GenerateTypeAdapter {
+
+  TypeAdapterFactory FACTORY = new TypeAdapterFactory() {
+    private final Map<Class<?>, Constructor<? extends TypeAdapter>> adapters =
```
JakeWharton: Needs to be `ConcurrentHashMap` since this isn't a thread-safe type

rharter:
Outside of Jake's comments and the failing tests this looks great!

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
hzsweers: as in `elements.stream.map(e -> Pair(e, getTypeAdapterMethod(e)).collect(toList())`?

JakeWharton: Yep!

hzsweers: I'll need to find a Pair implementation I can use here, oddly not even guava has it in its kitchen sink :|. Might just handwrite in a little holder here

auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
hzsweers: 06ac048

example/src/main/java/com/ryanharter/auto/value/gson/example/WebResponse.java
hzsweers: f09082c

example/src/main/java/com/ryanharter/auto/value/gson/example/WebResponse.java
hzsweers: bcdb966

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
hzsweers: e4b8b79

hzsweers:
Let me know if you want me to add a compile-testing test for this too (just did an integration test for now). I'll also add something to the README

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
JakeWharton: `AbstractMap.SimpleEntry`

hzsweers: Mind if I just keep the pair that I added here? Technically it doesn't really need to be a map anyway, I was just sheepishly using it to get an entry set 😁

README.md
```diff
@@ -138,6 +138,80 @@ android {
...
+new GsonBuilder()
+    .addTypeAdapterFactory(GenerateTypeAdapter.FACTORY)
+    .build()
+    .toJson(myFooInstance);
+```
+
+Proguard rules for obfuscation:
+
+```
+# Retain generated classes that end in the suffix
```
hzsweers: I _think_ this works, but going to ask a couple proguard-knowledgable friends to double check. We should only need to keep the names

rharter: I'm curious if this is even necessary.  This is a simple copy/paste, since the wildcards mean that nothing changes based on the implementation, and we now have a compile scoped artifact, so can we include the proguard config in the artifact the same way other libraries do?

rharter: Turns out the answer is no.  From Eric Lafortune:

> ProGuard consumer files are part of the aar specification, and implemented in the Android Gradle plugin (and the DexGuard Gradle plugin). I don't think there's a similar specification or implementation in the Java world.

So unless we make the dependency an aar, which I don't think we want to do, this is the way to go.

hzsweers: Yep. Agreed it doesn't make sense to make this an AAR as well.

PR #163, opened 2018-01-04
Type adapters are initialized lazily when needed rather than eagerly …

awelc:
This proposed change is motivated by the fact that in some cases not all fields of classes annotated with AutoValue and processed by auto-value-gson are used at the same time (or, in fact, used at all due to not being properly cleaned up by developers). Yet, in the the current solution, all type adapters representing these fields are initialized eagerly in the GsonTypeAdapter constructor forcing loading of all classes whether they are immediately useful or not. This particularly noticeable when these classes themselves are generated from (many different) specifications where developers clump together often unrelated fields.

More concretely, at Uber we have ~1700 classes that are autogenerated and then processed using auto-value-gson. Loading so many classes has implications on the application startup time - hence the idea of trying to reduce their numbers. With the proposed change, when the type adapters are not initialized until they are needed, we manage to reduce the number of autogenerated classes by ~30%, which has noticeable impact on startup performance.

It seems like the proposed change can help optimizing away class loading for similar scenarios without significantly compromising overall performance as the cost of accessing type adapters is in any case likely dwarfed by the cost or related I/O.


JakeWharton:
> With the proposed change, when the type adapters are not initialized until they are needed, we manage to reduce the number of autogenerated classes by ~30%, which has noticeable impact on startup performance.

The number of generated classes isn't changing so how are you reducing it?

> It seems like the proposed change can help optimizing away class loading for similar scenarios without significantly compromising overall performance as the cost of accessing type adapters is in any case likely dwarfed by the cost or related I/O.

Would still like to see numbers here, as you could write a reflection-based auto-value type adapter (now that the annotation has runtime retention) and eliminate all 1700 generated classes.

JakeWharton:
Also Gson is lazy by default so the only way these lookups would be costly is if you eagerly look up the type adapters needed for serialization. And if that's the case you can write a `LazyTypeAdapterFactory` which matches on all types, captures the `Gson` instance and `Type`, and then looks up the delegate only when `fromJson` or `toJson` are called.

rharter:
Thanks for the submission!  I like the idea of supporting lazily loaded adapters, but I'm not sure that this is the right approach.  AVG caches the property adapters to reduce the need for hash lookups that would have a runtime performance impact with Gson's built in caching.  While this eager lookup seems to be causing these load time issues due to your scale, it's still a runtime perf win.

It seems like the problem is that when a single TypeAdapter is initialized it eagerly initializes all property TypeAdapters, meaning that, potentially, your entire graph's TypeAdapters are being initialized all at once.

This approach throws the baby out with the bath water, though, by removing the benefit of cached adapters to reduce load time.  I think a better solution would be what @JakeWharton mentioned, having a lazy type adapter returned from the factory.  This removes the responsibility from the AutoValue extension, and moves it into the sibling processor for the TypeAdapterFactory.  

What I'm thinking is adding a property to the annotation so you can opt in with `@GsonTypeAdapterFactory(lazy = true)`, prompting the factory processor to wrap generated TypeAdapters with a `LazyTypeAdapter`, like so:

```java
public final class AutoValueGson_MyAdapterFactory extends MyAdapterFactory {

  @Override
  @SuppressWarnings("unchecked")
  public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
    Class<T> rawType = (Class<T>) type.getRawType();
    if (Foo.class.isAssignableFrom(rawType)) {
      return new LazyTypeAdapter<Foo>(gson, new DelegateFactory<Foo>() {
        public TypeAdapter<Foo> create(Gson gson) {
          return Foo.jsonAdapter(gson);
        }
      });
    } else if (Bar.class.isAssignableFrom(rawType)) {
      return new LazyTypeAdapter<Bar>(gson, new DelegateFactory<Bar>() {
        public TypeAdapter<Bar> create(Gson gson) {
          return Bar.jsonAdapter(gson);
        }
      });
    } else if (Baz.class.isAssignableFrom(rawType)) {
      return new LazyTypeAdapter<Baz>(gson, new DelegateFactory<Baz>() {
        public TypeAdapter<Baz> create(Gson gson) {
          return Baz.jsonAdapter(gson);
        }
      });
    } else {
      return null;
    }
  }

  private interface DelegateFactory<T> {
    TypeAdapter<T> create(Gson gson);
  }

  private static class LazyTypeAdapter<T> extends TypeAdapter<T> {
    
    private final Gson gson;
    private final DelegateFactory factory;
    private TypeAdapter<T> delegate;

    LazyTypeAdapter(Gson gson, DelegateFactory factory) {
      this.gson = gson;
      this.factory = factory;
    }

    private TypeAdapter<T> getDelegate() {
      if (delegate == null) {
        synchronized(this) {
          if (delegate == null) {
            delegate = factory.create(gson);
          }
        }
      }
      return delegate;
    }

    @Override public void write(JsonWriter out, T value) throws IOException {
      getDelegate().write(out, value);
    }

    @Override public List<String> read(JsonReader in) throws IOException { 
      return getDelegate().read(in);
    }
  }
}
```

This defers the loading of the TypeAdapter until it is needed (a huge win if you have lots of `@Nullable` fields), mitigating the init cost, but doesn't get rid of the benefit of caching the property TypeAdapters so they don't have to be looked up from Gson's hash.

awelc:
It does not seem like I can respond to individual comments so let me try addressing them in order.

I indeed misspoke, to an extent, when talking about class number reduction. The real intent there was to say that the number of classes observed to be loaded at startup is reduced by 30%. It is important to us, as the class loading time takes a significant portion of our startup time and we would like to defer loading of classes that are not strictly needed during startup until later time.

I am not sure if these are the numbers requested, but on Moto G 1st gen (we do care about older devices) running Android 7.1.2 we observed ~4.5 performance improvement (over 20 runs) of total cold application startup time, when comparing a version of the app built using "eager" (current) and "lazy" (proposed) version of AVG.

As far as I understand it, the `LazyTypeAdapterFactory` suggestion addresses a slightly different problem, at a different level of granularity. The laziness it is attempting to introduce is at the level of the factory while the problem we are trying to solve is at the level of individual generated type adapters (more details below). In fact, we do not have this particular factory problem as we generate our factory differently and instantiate adapters in the factory reflectively (which might have been one of Jake's suggestions).

Here is the problem we are facing. Consider the following class:

```
@AutoValue public abstract class Test {
  public abstract Bar a();
  public abstract Baz b();

  public static TypeAdapter<Test> typeAdapter(Gson gson) {
    return new AutoValue_Test.GsonTypeAdapter(gson);
  }
}
```

Further consider that `b` is optional here. Yet, in the generated code (in `GsonTypeAdapter` constructor), we will have:

```
public GsonTypeAdapter(Gson gson) {
  this.a = gson.getAdapter(Bar.class);
  this.b = gson.getAdapter(Baz.class);
}
```

This means that, whether `b` is actually used or not, both `Bar` and `Baz` classes will be loaded.

That being said, we did consider a solution somewhat similar to `LazyTypeAdapter`. We could generate accessors for `a` and `b` fields (possibly with some added synchronization):

```
private TypeAdapter<Bar> a() {
   if (a == null) a = gson.getAdapter(Bar.class);
   return a;
}

private TypeAdapter<Baz> b() {
   if (b == null) a = gson.getAdapter(Baz.class);
   return b;
}
```

We would then use the accessor methods instead of the field accesses in `read` and `write` methods.

This solution, however, is also problematic for us as it increases the number of methods which in turn has an impact on the binary size. Since we are already dealing with multiple dex files, adding thousands (or more, as some of our adapter classes are sizeable) of new methods is a no go...

I understand that you are concerned about the increased cost of the type adapter lookup. Still, the adapters are cached and looked up only in a context of an I/O operation - lookup cost should then be negligible compared to I/O cost.

rharter:
I'd like to check that I understand correctly.  The issue is not that the `TypeAdapter`s are being loaded, but rather that the classes are being loaded to fulfill the `Bar.class` portion of loading the TypeAdapter constructor.

If that is the case, would not a `LazyTypeAdapterFactory` solve this?  With the sample code I supplied, the constructor would never be called on a TypeAdapter until the TypeAdapter is needed (`read` or `write` are called).

Given the following example, when a `TypeAdapter` is requested for `Foo`, it returns a wrapped lazy type adapter. The ctor, which calls `gson.getAdapter(Bar.class)` isn't called until either `fooTypeAdapter.read()` or `fooTypeAdapter.write()` is called for the first time.  Even after it has been called, `gson.getAdapter(Baz.class)` won't be called until `barTypeAdapter` is used for the first time. This moves the class loading from all up from to at first time of use.

```java
@AutoValue abstract class Foo {
  public abstract Bar bar();
}

@AutoValue abstract class Bar {
  public abstract Baz baz();
}
```

This would result in a situation where the classloader only loads the root object and it's immediate properties.  Is that not enough, or not the right problem?

awelc:
I still think that our problem is at a finer granularity level, but clearly I may be missing something... I think that the `LazyTypeAdapterFactory` solution could work if `@AutoValue` was used on classes with only one abstract method to be converted into a type adapter, much like in the last example. We would have to use reflection quite cleverly there, otherwise the sheer fact of accessing the `class` field of a class would trigger class loading.

On the other hand, my previous example deals with classes that compound multiple type adapter initializations in a single class (currently in this class's constructor), and I am not sure how `LazyTypeAdapterFactory` would help here.

Let me extend my example to show how a fragment of autogenerated code for the `read`method of my `Test` class currently looks like:

```
@Override
public WebResponse<T> read(JsonReader jsonReader) throws IOException {
...
        switch (_name) {
          case "a": {
            status = a.read(jsonReader);
            break;
          }
          case "b": {
            data = b.read(jsonReader);
            break;
          }
          default: {
            jsonReader.skipValue();
          }
        }
...
}
```

Let's then apply `LazyTypeAdapterFactory` to my `Test` class:

```
public final class AutoValueGson_MyAdapterFactory extends MyAdapterFactory {

@Override
@SuppressWarnings("unchecked")
public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
  Class<T> rawType = (Class<T>) type.getRawType();
  if (Test.class.isAssignableFrom(rawType)) {
    return new LazyTypeAdapter<Test>(gson, new DelegateFactory<Test>() {
      public TypeAdapter<Test> create(Gson gson) {
        return Test.jsonAdapter(gson);
      }
    });
  } else {
    return null;
  }
}
```

The issue here is that, indeed, the constructor of `Test` class will only be called when it's needed by, say, `read` method, but once it's called, it will still force loading of `Bar` and `Baz` classes that `Test` references, even though `Baz` may not be needed due to `b` being optional in the incoming JSON message. In the proposed modification we don't have `b` field in the generated code so if the specific case of the switch statement is not used, then a given class is never loaded.

awelc:
I have created a new version that (hopefully) alleviates most of your concerns (the adapters are still cached in fields, but these fields are initialized lazily without increasing the number of existing methods). I will create a new branch (lazy_type_adapters_with_fields) and a new pull request containing the new version.

PR #165, opened 2018-01-08
Type adapters are initialized lazily when needed rather than eagerly …

awelc:
Here is another take on the modification moving type adapter initialization from eager (in the constructor) to lazy (initialized only when needed). 

Please note that I intentionally left out updating correctness tests - the plan is to fix them after reaching consensus on the proposed changes.

The gist of the proposed update is to keep the fields to cache initialized type adapters, but instead of initializing these adapter caching fields eagerly in the constructor, they are initialized when a particular field is serialized/deserialized.

The type adapter caching fields are now volatile which should guarantee thread safety as the volatile field updates are ordered (and atomic), and threads racing to update these fields will update them with the same value, so it does not matter which one succeeds first (or even if more than one threads manages to make the update).

Modification of the code to generate the `read` method was straightforward - in each case of the `switch` statement we now check if the adapter caching field has been initialized and then initialize it if it wasn't.

The write methods has been modified a bit more to take advantage of the lazy nature of the adapter caching fields. Previously, regardless of whether a given value field was initialized or not, a special type adapter was created to serialize this field (even if the value of the field was `null`). With lazy adapters, we can do better than that - for non-primitive value fields we can check if the value to be serialized is `null` and initialize the adapter only if it is not. We could do the same for primitive values as well but as the focus of this optimization is to reduce the overhead related to initializing custom adapters (e.g. overhead related to loading classes required to initialize adapters), it's unlikely that handling primitive types specially would be highly beneficial.

For reference, I am attaching two sample files (from the AVG project examples) - one representing code generated by the "original" (unmodified) AVG version (AutoValue_WebResponseOrg.txt) and the other one representing code generated by the proposed AVG version (AutoValue_WebResponse.txt). I used .txt extension as GitHub apparently does not allow attaching .java files.

[AutoValue_WebResponse.txt](https://github.com/rharter/auto-value-gson/files/1613005/AutoValue_WebResponse.txt)
[AutoValue_WebResponseOrg.txt](https://github.com/rharter/auto-value-gson/files/1613006/AutoValue_WebResponseOrg.txt)


rharter:
This looks good to me.  While it's a bit messy, I think having the null checks inlined isn't an issue for generated code and reduces the generated methods. @JakeWharton any thoughts?

rharter:
I think this is good to move forward with tests, etc.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -349,7 +350,7 @@ public String generateClass(Context context, String className, String classToExt
       if (!seenTypes.contains(property.type)) {
         fields.put(property.type,
                 FieldSpec.builder(adp,
-                    nameAllocator.newName(simpleName(property.type)) + "_adapter", PRIVATE, FINAL)
+                                  nameAllocator.newName(simpleName(property.type)) + "_adapter", PRIVATE, VOLATILE)
```
rharter: nit: The current convention in this project is for parameter continuation to have two tabs (8 spaces), not align with other parameters.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -554,9 +540,61 @@ public TypeSpec createTypeAdapter(Context context,
...
+                                          Types typeUtils,
+                                          List<TypeVariableName> typeParams) {
+    TypeName type = prop.type.isPrimitive() ? prop.type.box() : prop.type;
+    ParameterizedTypeName adp = ParameterizedTypeName.get(jsonAdapter, type);
+    CodeBlock.Builder codeBuilder = CodeBlock.builder();      
+
+    if (prop.typeAdapter != null) {
+      if (typeUtils.isAssignable(prop.typeAdapter, typeAdapterFactory)) {
+        if (prop.type instanceof ParameterizedTypeName || prop.type instanceof TypeVariableName) {
+          codeBuilder.add("(($T) new $T().create(gson, $L))", adp,
```
rharter: Can we ditch the extra parenthesis? They make the generated code a bit more complex to read without adding value.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
awelc: My apologies, I didn't notice. Will fix!

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
awelc: This is an artifact of the previous version where there were no fields and generated sequence was used directly as a method receiver rather than being RHS of the assignment. Will remove in the next revision.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
rharter: You know what, I was wrong, the project uses 2 space indentation, so it's actually 4 spaces for the continuation.  To match the `.build()` line.  Sorry for the mixup.

awelc: I just pushed another change. Hopefully I got this one right - I put 4 spaces for both the parameter continuation indentation and for the following line starting with ".".

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -178,35 +171,136 @@ public void simple() {
...
         + "      jsonWriter.beginObject();\n"
         + "      jsonWriter.name(\"a\");\n"
-        + "      string_adapter.write(jsonWriter, object.a());\n"
+        + "      if (object.a() == null) {\n"
+        + "        jsonWriter.nullValue();\n"
+        + "      } else {\n"
+        + "        if (string_adapter == null) {\n"
+        + "          string_adapter = gson.getAdapter(java.lang.String.class);\n"
+        + "        }\n"
+        + "        string_adapter.write(jsonWriter, object.a());\n"
```
JakeWharton: I don't like that this creates 3 volatile reads in a row. I would prefer if the pattern
```java
AdapterType string_adapter = this.string_adapter;
if (string_adapter == null) {
  this.string_adapter = string_adapter = gson.getAdapter...
}
string_adapter.write(..)
```
which only requires a single volatile read.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -178,35 +171,136 @@ public void simple() {
...
+        + "        }\n"
+        + "        string_adapter.write(jsonWriter, object.d());\n"
+        + "      }\n"
         + "      jsonWriter.name(\"e\");\n"
-        + "      string_adapter.write(jsonWriter, object.e());\n"
+        + "      if (object.e() == null) {\n"
+        + "        jsonWriter.nullValue();\n"
+        + "      } else {\n"
+        + "        if (string_adapter == null) {\n"
+        + "          string_adapter = gson.getAdapter(java.lang.String.class);\n"
```
JakeWharton: Why are the FQCNs of these types being used instead of imports here? You're likely calling `toString()` or using `$L` with a `TypeName` instead of using `$T`.

JakeWharton:
Still not a fan of this change. Feels like your bad modeling design leaking out into your dependencies.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
awelc: Good point. I wonder what the performance effect would be (these operations are unlikely to be contended which I believe increases the cost) but reducing number of volatile reads (I think only, by one, though) will certainly not hurt. Will fix in a separate PR.

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
awelc: I totally agree - FQCNs are unnecessary. Will fix!

PR #167, opened 2018-01-16
Switch from ConcurrentHashMap to Collections.synchronizedMap() in GenerateTypeAdapter

hzsweers:
ConcurrentHashMap bugs on some versions of android: https://code.google.com/p/android/issues/detail?id=170073

### DIFF
--- a/auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
+++ b/auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
@@ -9,8 +9,9 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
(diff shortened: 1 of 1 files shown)


rharter:
Got some test failures.

hzsweers:
This is what I get for editing the file directly on GitHub :)

auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
```diff
@@ -27,7 +28,7 @@
 
   TypeAdapterFactory FACTORY = new TypeAdapterFactory() {
     private final Map<Class<?>, Constructor<? extends TypeAdapter>> adapters =
-        new ConcurrentHashMap<>();
+        Collections.synchronizedMap(new HashMap<Class<?>, Constructor<? extends TypeAdapter>>());
```
NightlyNexus: btw, [LinkedHashMap is always better than HashMap.](https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/)

hzsweers: Ordering doesn't matter here nor are we testing it, I'm not too worried about it :)

hzsweers:
We have an error-prone check against it internally, but yeah seems a good candidate

auto-value-gson-annotations/src/main/java/com/ryanharter/auto/value/gson/GenerateTypeAdapter.java
JakeWharton: Ordering _always_ matters regardless on whether you explicitly rely on it or not. That's the point of the post.

hzsweers: I don't feel strongly about it in this case, but done in 24fc3b7

hzsweers:
reran travis a couple times, seems it's been flaky the past few days. Green now

PR #168, opened 2018-01-24
Lazy type adapters fixes

awelc:
This implements both of Jake's suggestions for the previous pull request on lazy type adapters - remove unnecessary full qualification from type names and adopt a pattern that reduces the number of volatile reads.


JakeWharton:
Thanks for following up!

awelc:
More than happy to :-)

rharter:
Great!  Thanks for adding these!

PR #175, opened 2018-04-12
Refactor code that was using Java 8 features

jakekdodd:
Broken in #157 

(sorry for the couple of automatic whitespace changes...I can resubmit without those if you wish)

### DIFF
--- a/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
+++ b/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
@@ -18,6 +18,7 @@ import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Why? AutoValue's processor requires Java 8 so this isn't going to prevent the requirement to use JDK 8 in your builds.

jakekdodd:
@JakeWharton #157. I didn't read through the [whole conversation](https://github.com/rharter/auto-value-gson/issues/142#issuecomment-349002607) but you're welcome to do so.

The build on `master` is [failing](https://travis-ci.org/rharter/auto-value-gson/builds/356921700) (regardless of whether you build with JDK 7 or 8—I haven't touched 7 in years) and the fix only took a minute, so I did it. Since AutoValue's processor requires JDK 8 anyways, we should probably just revert the commit(s) that busted the build

NightlyNexus:
isn't this change the culprit?
https://github.com/rharter/auto-value-gson/pull/145
why did the build for it succeed?

jakekdodd:
@NightlyNexus the CI build happened last August when that PR was first submitted, which was before the Java 8 source was merged

NightlyNexus:
ahaha. good eye. perhaps Ryan would accept a PR to revert that.

PR #176, opened 2018-04-12
Revert Java 7 compatibility

jakekdodd:
@JakeWharton I think these are all of the Java 7 compatibility commits =)

### DIFF
--- a/auto-value-gson/build.gradle
+++ b/auto-value-gson/build.gradle
@@ -3,8 +3,8 @@ import org.gradle.internal.jvm.Jvm
 apply plugin: 'java-library'
 apply plugin: 'maven-publish'
 
-sourceCompatibility = JavaVersion.VERSION_1_7
-targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
 
 dependencies {
     api project(':auto-value-gson-annotations')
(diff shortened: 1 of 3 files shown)


hzsweers:
These are for runtime and packaging use

the runtime adapter and annotations are java 7 so they can be embedded in android environments without requiring minSdk 24. Jake's comment in the other PR was regarding requirements during annotation processing, not runtime.

jakekdodd:
@hzsweers something needs to change or this project will not compile. I can either (a) reopen the original pull request or (b) resubmit this pull request with the Java 7 compatibility limited to the annotations module

auto-value-gson/build.gradle
```diff
@@ -3,8 +3,8 @@ import org.gradle.internal.jvm.Jvm
 apply plugin: 'java-library'
 apply plugin: 'maven-publish'
 
-sourceCompatibility = JavaVersion.VERSION_1_7
-targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
```
NightlyNexus: this is the only change that needs to happen. the rest are not necessary.

hzsweers: Oops, missed that part. Yes this part can be reverted. Rest should stay java 7 though for the reasons listed above

hzsweers: Note I can't reopen the PR because github says your branch was "force pushed or recreated"

PR #177, opened 2018-04-13
Revert "Add compatilibity to java 1.7"

jakekdodd:
This reverts commit 2a7275f78aee0b5bb49fcfe88274ee2831e65014.

Only auto-value-gson-annotations needs to be JDK7-compatible.


jakekdodd:
@hzsweers I did force push to that old branch so here's a new one

PR #185, opened 2018-08-23
Adding support for optional types

kaibolay:
This is a first-cut implementation for issue #123, adding support for
the following optional types:

com.google.common.base.Optional
java.util.Optional
java.util.OptionalDouble
java.util.OptionalInt
java.util.OptionalLong

Regression tests have been added to cover various combinations of features.

This is based on https://github.com/rharter/auto-value-gson/pull/148 by
Alejandro J. C De Baca <acdebaca@github> and fixes merge conflicts.


acdebaca:
@kaibolay Thanks for doing this. I recall there being one change that Ryan and I discussed in the PR that would be necessary to get this PR merged regarding how to default nullable `Optional`s. I'll take a look.

hzsweers:
I think we should make a call for now to just not support these. Modeling absence in general Java code is one thing, but modeling it in JSON is very different, and I'd rather opt for no solution rather than a limited one and encourage better support for normalization from AutoValue via #132.

Issue #164, opened 2018-01-07
Jdk9 Support


kartoffelsup:
First of all, thanks for this nice project :) 

When trying to use `auto-value-gson` with jdk9 the following errors occur:
using:
```gradle
  compile('com.ryanharter.auto.value:auto-value-gson-annotations:0.7.0')
  apt('com.ryanharter.auto.value:auto-value-gson:0.7.0')
```
error:
```
error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider com.ryanharter.auto.value.gson.AutoValueGsonAdapterFactoryProcessor could not be instantiated
```
which can be resolved with:
```
  compile('com.ryanharter.auto.value:auto-value-gson:0.7.0')
  compile('com.ryanharter.auto.value:auto-value-gson-annotations:0.7.0')
  apt('com.ryanharter.auto.value:auto-value-gson:0.7.0')
```
but leads to more errors:
```
error: the unnamed module reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module auto.value.gson reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module auto.common reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module javapoet reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module guava reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module gson reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module auto.value.gson.annotations reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
error: module auto.value reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
F:\dev\auto-value-gson-jdk9\src\main\java\module-info.java:1: error: module com.example.gson.auto.value reads package com.ryanharter.auto.value.gson from both auto.value.gson.annotations and auto.value.gson
module com.example.gson.auto.value {
^
9 errors

```
repro:
[auto-value-gson-jdk9.zip](https://github.com/rharter/auto-value-gson/files/1609605/auto-value-gson-jdk9.zip)

kartoffelsup:
[...]

PR #186, opened 2018-10-08
#164 Use correct "Generated"-annotation based on JDK version.

ChristianCiach:
Fixes usage of auto-value-gson with JDK 9 and newer.

This bumps the dependency of auto-common to 0.9, because this version includes the necessary utility method.

### DIFF
--- a/auto-value-gson/build.gradle
+++ b/auto-value-gson/build.gradle
@@ -13,7 +13,7 @@ dependencies {
     api 'com.google.auto.value:auto-value:1.6'
     api 'com.google.auto.value:auto-value-annotations:1.6'
     compileOnly 'com.google.auto.service:auto-service:1.0-rc3'
-    api 'com.google.auto:auto-common:0.8'
+    api 'com.google.auto:auto-common:0.9'
 
     testCompile 'junit:junit:4.12'
     testCompile 'com.google.truth:truth:0.36'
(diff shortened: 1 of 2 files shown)


geminicaprograms:
Hi,
Any idea when it gets released?

Thanks
Jacek

hzsweers:
When it's ready

ChristianCiach:
I second that request. The fix is urgent for anyone using Java 9 and above. Never has there been such a long gap between releases.

hzsweers:
Urgent needs can always build from source if they need to. Let’s avoid the dramatics in the issue tracker

geminicaprograms:
Likewise, simple answer hinting when it can be expected or what is blocking it would do but you decided differently so you got dramatics :)

rharter:
I'm going to lock conversation on this merged pull request since none of this is helpful.

I understand the frustration in having to wait for changes to be released, however there is someone waiting for every change and we decide when to cut releases based on work planned.

We have some more bugs and issues to fix up before this is released, as we'd like to do a single release that includes fixes for many of the issues that have grown here, but we won't provide a timeline because this is something we're doing in free time and, especially this time of year, we literally don't know how long things are going to take as we also have families and other obligations to attend.

As @hzsweers mentioned, you're welcome to build from source if your needs are urgent, otherwise we're going to finish up the issues we can and then will cut a release.

Issue #48, opened 2016-06-05
Ignore(BOTH) isn't respected


shalzz:
Provide an annotation that lets us specify the fields _not_ to be included in the generated TypeAdapter for the `@AutoValue` annotated class.

For Example:

``` java
@AutoValue public abstract class Foo {
  @Gson.Ignore abstract String bar();
  @SerializedName("Baz") abstract String baz();

  // The public static method returning a TypeAdapter<Foo> is what
  // tells auto-value-gson to create a TypeAdapter for Foo.
  public static TypeAdapter<Foo> typeAdapter(Gson gson) {
    return new AutoValue_Foo.GsonTypeAdapter(gson);
  }
}
```

Generated Class:

``` java
public static final class GsonTypeAdapter extends TypeAdapter<Foo> {
    private final TypeAdapter<String> bazAdapter;
    public GsonTypeAdapter(Gson gson) {
      this.bazAdapter = gson.getAdapter(String.class);
    }

   // ...
}
```

Shalzz:
Just saw your comment in #27 , keep this open till support for it is added?

eliasbagley:
What's the status on this? I need to be able to ignore certain fields from being serialized

aegis123:
you can use `@Ignore(SERIALIZATION)`, `@Ignore(DESERIALIZATION)` or `@Ignore(BOTH)` to ignore abstract methods from typeadapter creation.

ubuntudroid:
@aegis123 Unfortunately BOTH doesn't work as it isn't checked for.

aegis123:
@ubuntudroid sorry my mistake `BOTH` is the default option for `@Ignore` so you don't have to specify it.

hzsweers:
I cannot repro this, so feel free to re-open with a repro code example. The test for this is pretty comprehensive as well. I am making some minor cleanups to the logic in #196

PR #196, opened 2019-01-13
Clean up Ignore logic a bit

ZacSweers:
Relates to #48, but I can't repro any issue described there and think it may just be old. This improves it a bit by not always inverting the serialization checks.

I do plan to just use a shared transient annotation library for this in a separate PR, but this should make switching to that easy

### DIFF
--- a/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -158,14 +158,14 @@ public class AutoValueGsonExtension extends AutoValueExtension {
       return serializedName != null;
     }
 
-    public boolean shouldSerialize() {
+    public boolean shouldNotSerialize() {
       Ignore ignore = element.getAnnotation(Ignore.class);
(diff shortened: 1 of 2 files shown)


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -358,7 +358,7 @@ private static AnnotationSpec createGeneratedAnnotationSpec(TypeElement generate
     Set<TypeName> seenTypes = Sets.newHashSet();
     NameAllocator nameAllocator = new NameAllocator();
     for (Property property : properties) {
-      if (!property.shouldDeserialize() && !property.shouldSerialize()) {
+      if (property.shouldNotDeserialize() && !property.shouldNotSerialize()) {
```
rharter: Should this be `property.shouldNotSerialize()` (drop the `!`)?

ZacSweers: Derp, will fix

ZacSweers: 2dd551a

PR #202, opened 2019-01-15
Refactory annotations to runtime artifact

### LINKED ISSUE #114: List<? extends MyType> fails since 0.4.6
If you add a following method it fails in the new version (passes with 0.4.5)
public abstract List<? extends MyType> bar();

```
../AutoValue_Foo.java:117: error: illegal start of expression
      this.fooAdapter = (TypeAdapter<List<? extends MyType>>) gson.getAdapter(TypeToken.getParameterized(List.class, ? extends MyType.class));
                                                                                                                                         ^
.../AutoValue_Foo.java:117: error: illegal start of expression
      this.fooAdapter = (TypeAdapter<List<? extends MyType>>) gson.getAdapter(TypeToken.getParameterized(List.class, ? extends MyType.class));
                                                                                                                                           ^
2 errors
```

ZacSweers:
This refactors the annotations artifact to be a full runtime artifact, and also adds support for wildcards via the introduction of `Types`

Resolves #114


auto-value-gson-runtime/src/main/java/com/ryanharter/auto/value/gson/internal/Util.java
```diff
@@ -0,0 +1,226 @@
...
+package com.ryanharter.auto.value.gson.internal;
+
+import java.lang.reflect.GenericArrayType;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
+import java.util.Arrays;
+import org.jetbrains.annotations.Nullable;
+
+final class Util {
```
rharter: I realize this is copied in, but would you mind giving it a more useful name like `TypeUtil`?

ZacSweers: 6c5f780

PR #203, opened 2019-01-15
Embed proguard rules

### LINKED ISSUE #200: Embed proguard rules in the runtime artifact
From the README

```
# Retain generated classes that end in the suffix
-keepnames class **_GsonTypeAdapter

# Prevent obfuscation of types which use @GenerateTypeAdapter since the simple name
# is used to reflectively look up the generated adapter.
-keepnames @com.ryanharter.auto.value.gson.GenerateTypeAdapter class *
```

We can put these in the runtime artifact. We can also improve them to use if-rules where possible like Moshi does

ZacSweers:
This embeds proguard rules directly in the jar for R8/proguard, and also tunes them to use newer features to only selectively keep adapters (adapted from Moshi's configuration)

Resolves #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,9 +187,9 @@ new GsonBuilder()
 
 ### R8 / ProGuard
 
-If you are using R8 or ProGuard add the options from [this file](https://github.com/rharter/auto-value-gson/blob/master/auto-value-gson-runtime/src/main/resources/META-INF/proguard/autovaluegson.pro). If using Android, this requires Android Gradle Plugin 3.2.0+.
+If you are using R8 or ProGuard add the options from [this file](https://github.com/rharter/auto-value-gson/blob/master/auto-value-gson-runtime/src/main/resources/META-INF/proguard/autovaluegson.pro).
 
-On Proguard 6.1+ or R8, these rules will be automatically detected.
+If using Android, Android Gradle Plugin 3.3.0+, these rules will automatically be detected.
 
 ## Download
 
--- a/README.md
+++ b/README.md
@@ -185,16 +185,13 @@ new GsonBuilder()
     .toJson(myFooInstance);
 ```
 
-Proguard rules for obfuscation:
 
-```
-# Retain generated classes that end in the suffix
--keepnames class **_GsonTypeAdapter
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -185,16 +185,13 @@ new GsonBuilder()
...
-# Retain generated classes that end in the suffix
--keepnames class **_GsonTypeAdapter
 
-# Prevent obfuscation of types which use @GenerateTypeAdapter since the simple name
-# is used to reflectively look up the generated adapter.
--keepnames @com.ryanharter.auto.value.gson.GenerateTypeAdapter class *
-```
+### R8 / ProGuard:
+
+If you are using R8 or ProGuard add the options from [this file](https://github.com/rharter/auto-value-gson/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro). If using Android, this requires Android Gradle Plugin 3.2.0+.
```
rharter: Is this link right?

ZacSweers: ecce2db

README.md
```diff
@@ -185,16 +185,13 @@ new GsonBuilder()
...
+
+On Proguard 6.1+ or R8, these rules will be automatically detected.
```
gabrielittner: The automatic rule detection is part of AGP 3.3.0 not of proguard 6.1

README.md
```diff
@@ -185,34 +185,23 @@ new GsonBuilder()
     .toJson(myFooInstance);
 ```
 
-Proguard rules for obfuscation:
+### R8 / ProGuard
 
-```
-# Retain generated classes that end in the suffix
--keepnames class **_GsonTypeAdapter
+If you are using R8 or ProGuard add the options from [this file](https://github.com/rharter/auto-value-gson/blob/master/moshi/src/main/resources/META-INF/proguard/autovaluegson.pro). If using Android, this requires Android Gradle Plugin 3.2.0+.
```
gabrielittner: There's still one moshi in the link ;)

ZacSweers: 85a6669 :|

README.md
ZacSweers: 99054c1

PR #204, opened 2019-01-16
 Remove invalid runtime in gradle script.

TonyTangAndroid:


TonyTangAndroid:
I see. Because currently, if I use 
```compile 'com.ryanharter.auto.value:auto-value-gson-runtime:0.8.0'```, 
it will NOT be found and gradle build fails. 

However, if I use 
```compile 'com.ryanharter.auto.value:auto-value-gson:0.8.0'```, 
it works.

PR #209, opened 2019-01-28
Extract field renaming util to runtime artifact

### LINKED ISSUE #206: Move field naming support to runtime artifact
These are duplicated for each model right now and a bit wasteful

ZacSweers:
WIP, will update tests once #207 merges

Resolves #206


ZacSweers:
I've got concerns about the amount of overhead this support brings in, and think we should make it opt-in via option. Will punt that to whenever autovalue releases again with support for plugins declaring custom supported options though. Some of the generated code for field names are also a bit odd

PR #210, opened 2019-02-01
Full support for non-public typeAdapter() methods

### LINKED ISSUE #205: Support package-private typeAdapter methods
Analogous to https://github.com/rharter/auto-value-moshi/issues/87

ZacSweers:
Resolves #205


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
```diff
@@ -121,8 +125,43 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
-      TypeSpec typeAdapterFactory = createTypeAdapterFactory(type, elements, packageName,
+      PackageElement packageElement = packageElementOf(type);
+      String packageName = packageElement.getQualifiedName().toString();
+      List<TypeElement> applicableElements = elements.stream()
+          .filter(e -> {
+            Visibility typeVisibility = Visibility.ofElement(e);
+            switch (typeVisibility) {
+              case PRIVATE:
+                return false;
+              case DEFAULT:
```
rharter: Won't this fail to generate implementations for public elements from different packages? I think the package should only have to match if the element is protected, no?

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessorTest.java
```diff
@@ -45,6 +45,42 @@
             + "  }\n"
             + "  abstract String getName();\n"
             + "}");
+    // This is generated into a different package and not visible to the factory
```
rharter: This could use a test for types that are visible but in a different package (which I think will currently fail, based on my other comments).

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -487,6 +491,48 @@ public void simple() {
...
+        + "import com.google.gson.TypeAdapter;\n"
+        + "import com.google.gson.Gson;\n"
+        + "import java.util.Map;\n"
+        + "import java.util.Set;\n"
+        + "@AutoValue abstract class Test {\n"
+        + "  private static TypeAdapter<Test> typeAdapter(Gson gson) {\n"
+        + "    return null;\n"
+        + "  }\n"
+        // Reference type
+        + "public abstract String a();\n"
```
rharter: nit: these items could use a couple spaces to match formatting.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonAdapterFactoryProcessor.java
rharter: Ahh, okay, that sounds good then.

PR #215, opened 2019-04-26
Migrate off deprecated generatedAnnotation method

cushon:


ZacSweers:
Thanks! FYI that this project is currently not planning to see any new releases, but let us know if that's something you're interested in

cushon:
Thanks for the FYI. We're building it from source so not having an official release is fine. Is the future of the project uncertain, or you just don't have much time to cut releases?

ZacSweers:
We're trying to coalesce all of auto-value-gson/moshi/parcel to a stable
and feature-complete 1.0 and basically call it at that. Ryan and I both
don't have much time to put towards them and have both also essentially
moved on to kotlin-oriented solutions in our own work (Moshi codegen,
Parcelize)

PR #216, opened 2019-04-26
Add @RunWith annotations to JUnit tests

cushon:


ZacSweers:
Thanks! Out of curiosity - what spurred this? I've always wondered if it was necessary given you could run tests without them too

cushon:
> Thanks! Out of curiosity - what spurred this? I've always wondered if it was necessary given you could run tests without them too

As I understand it it's optional with the standard testrunners but still a good practice. We have a custom testrunner that is pickier about the annotations.

ZacSweers:
👍

Issue #158, opened 2017-12-04
Issue with default empty collections still being null


cammace:
I'm currently in the middle of upgrading one of our pure Java libraries using AutoValue and this extension. I'm running into an issue however while trying to set the compiler options in my submodules `build.gradle` I see no errors but it doesn't seem to change anything. 

My understanding of the default collections to empty is that any list will never be null but rather just an empty list? This isn't working properly for me right now and the list still ends up being null.

```groovy
compileJava {
  options.compilerArgs << "-Aautovaluegson.defaultCollectionsToEmpty=true"
}
```

When I try setting the list a default value using:

```groovy
options.compilerArgs << "-Aautovaluegson.mutableAdaptersWithDefaultSetters=true"
```

I just get an error that there is no such method for `setDefaultVoiceInstructions`

```java
public static TypeAdapter<LegStep> typeAdapter(Gson gson) {
    return new AutoValue_LegStep.GsonTypeAdapter(gson)
      .setDefaultVoiceInstructions(new ArrayList<VoiceInstructions>());
  }
```

Am I missing something here or does this functionality not work properly anymore? 

For reference, [this is the class](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/LegStep.java#L185) I am trying to implement this in, inside [this project](https://github.com/mapbox/mapbox-java).

hzsweers:
I think we should make a call for now to just not support these. Modeling absence in general Java code is one thing, but modeling it in JSON is very different, and I'd rather opt for no solution rather than a limited/awkward one and encourage better support for normalization from AutoValue via #132.

PR #217, opened 2019-04-27
Optionally initialize known collection types to empty, instead of null

cushon:
Context:
* https://github.com/rharter/auto-value-gson/commit/15564443795bdf3f0351db41d6de770cc2ceba1d
* https://github.com/rharter/auto-value-gson/commit/283f041c74cb740b94fe9ee8538dca36d13b9ac9

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,12 @@ Gson gson = new GsonBuilder()
     .create();
 ```
 
+## Compiler options
+
+`autovaluegson.defaultCollectionsToEmpty` - If specified, maps/collections will
+default to their empty types (e.g. `List` -> `Collections.emptyList()`). Value
+is `true` or `false`.
+
 ## @GenerateTypeAdapter
 
 There is an annotation in the `auto-value-gson-runtime` artifact called `@GenerateTypeAdapter`. This annotation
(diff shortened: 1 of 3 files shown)


cushon:
Would you be willing to consider restoring support for this feature? Its removal is preventing us from updating to head. I understand from #158 this may not have been a complete solution, but removing it without a replacement feels like a step backwards.

ZacSweers:
#136 is the ideal replacement, but is blocked by https://github.com/google/auto/pull/501 and something that Google does have control over. Not sure how connected you are with that project, but what would you think of revisiting that? I think if that API were finally reviewed, we'd be happy to do a new release with support for it and support property normalization like this indirectly

rharter:
Sorry for the delay on this, I've been on parental leave so haven't been doing much in the projects.

IIRC the problem with #158 was largely in the default values, as opposed to the empty lists. Would it be helpful to have empty list support without default values? It seems like that would be a relatively benign change, which I'd be okay with, but it's also largely a modelling issue (e.g. if the property should be an empty list, make it an empty list in json).

cushon:
> #136 is the ideal replacement, but is blocked by google/auto#501 and something that Google does have control over. Not sure how connected you are with that project, but what would you think of revisiting that?

It sounds like there's still interest in that PR, but the change needs to go through our API review proceses and I don't have an ETA for that yet.

> Would it be helpful to have empty list support without default values?

It would be, because we started depending on this handling of empty lists after 15564443795bdf3f0351db41d6de770cc2ceba1d was added, and the removal of that support is making it hard to upgrade to the latest version of auto-value-gson.

We investigated modifying the affected uses of auto-value-gson to use explicit empty lists instead of relying on this support, but it would be an invasive cleanup and it isn't clear it's an improvement for those use-cases.

ZacSweers:
Going to close this now that #136 is available. The solution for this would be to create a builder factory that defaults collections to empty. Note that if the collection is null in JSON, it would be overwritten to null in the builder!

Issue #161, opened 2017-12-27
Support for Optional field


dthong:
Optional fields are default to `Optional.empty()` in AutoValue, but I got null exception in deserialization. For those Optional fields, should the plugin initialize them to `Optional.empty()` instead of null before passing values from jsonReader?

kaibolay:
This is a duplicate of issue #123 - PR #148 fixes the issue, but has been pending for a while...

hzsweers:
I think we should make a call for now to just not support these. Modeling absence in general Java code is one thing, but modeling it in JSON is very different, and I'd rather opt for no solution rather than a limited one and encourage better support for normalization from AutoValue via #132.

surlyengineer:
FWIW I hear the point about modeling absence in JSON but this feature would be pretty handy.

PR #224, opened 2019-10-07
Support property defaults, normalization, and validation via Builders, Take 2

bryanstern:
This is a second pass at implementing a solution for #132 using the proposed AutoValue builder properties add in auto-value 1.7rc1 (see https://github.com/google/auto/issues/421). I decided to reimplement this rather than update #136 since it was so far behind master.

Unlike the original PR, `Optional` fields are not supported since that appears to be the direction this project has taken. See #161.

If an AutoValue class specifies a builder with the @AutoValue.Builder annotation, the generated `TypeAdapter#read` implementation will make use of the builder. If there is a builder factory method, it will be used. If there are multiple builder factory methods, the `@AutoValueGsonBuilder` will distinguish which one to use.


auto-value-gson-runtime/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonBuilder.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * If present, indicates that the annotated method should be used for retrieving an instance of the
+ * AutoValue.Builder. Only necessary if there is more than one builder method.
+ */
+@Retention(CLASS)
+@Target(METHOD)
+public @interface AutoValueGsonBuilder {
+}
```
ZacSweers:
```suggestion
}

```

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
     return readMethod.build();
   }
 
-  /** Returns a default value for initializing well-known types, or else {@code null}. */
+  private MethodSpec createReadMethod(ClassName className,
```
ZacSweers: Let's maybe rename this one to "createReadMethodWithBuilder" for clarity?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
...
+            .collect(Collectors.toSet());
+
+        if (annotatedMethods.size() == 0) {
+          processingEnvironment.getMessager()
+              .printMessage(
+                  Diagnostic.Kind.ERROR,
+                  "Too many builder methods in " + className.simpleName() + ". Annotate the "
+                      + "builder this TypeAdapter should use with @AutoValueGsonBuilder."
+              );
+          throw new IllegalStateException();
```
ZacSweers: We should just try to return early and skip generation in this case. Logging the error will defer to the compiler to report issues (it will still fail the build, but with _all_ relevant errors)

ZacSweers: Same with the below case

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
...
+              .printMessage(
+                  Diagnostic.Kind.ERROR,
+                  "Too many @AutoValueGsonBuilder annotations",
+                  annotatedMethods.stream().findFirst().get()
+              );
+          throw new IllegalStateException();
+        }
+      }
+
+      readMethod.addStatement("$T $N = $T.$L", builder.type, builder, autoValueClassName, builderMethod);
```
ZacSweers: Does JavaPoet automatically turn an ExecutableElement into its name + `()` here for `%L`?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
...
+    readMethod.addStatement("$N.nextNull()", jsonReader);
+    readMethod.addStatement("return null");
+    readMethod.endControlFlow();
+
+
+    // Instantiate the AutoValue'd class builder
+    FieldSpec builder = FieldSpec.builder(ClassName.get(builderContext.builderType()), "builder")
+        .build();
+
+    Set<ExecutableElement> builderMethods = builderContext.builderMethods();
```
ZacSweers: How do we handle parameters? A common pattern for builder factory methods is to accept required properties as its parameters.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
...
+    FieldSpec name = FieldSpec.builder(String.class, "_name").build();
+    readMethod.addStatement("$T $N = $N.nextName()", name.type, name, jsonReader);
+
+    readMethod.beginControlFlow("if ($N.peek() == $T.NULL)", jsonReader, token);
+    readMethod.addStatement("$N.nextNull()", jsonReader);
+    readMethod.addStatement("continue");
+    readMethod.endControlFlow();
+
+    readMethod.beginControlFlow("switch ($N)", name);
+    for (Property prop : properties) {
```
ZacSweers: A lot of this is duplicated. Could we maybe share this and just have a notion of "assignment" codeblocks that either set the builder property or assign to the local variable if there's no builder?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -671,7 +708,166 @@ private MethodSpec createReadMethod(ClassName className,
...
+    readMethod.addStatement("$N.skipValue()", jsonReader);
+    readMethod.endControlFlow(); // default case
+
+    readMethod.endControlFlow(); // switch
+    readMethod.endControlFlow(); // while
+
+    readMethod.addStatement("$N.endObject()", jsonReader);
+
+    // find the build method to use
+    if (!builderContext.buildMethod().isPresent()) {
```
ZacSweers: Let's find this eagerly before the property code above so we can fail fast

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -834,93 +809,1294 @@ public void simple() {
...
+    assertAbout(javaSources())
+        .that(Arrays.asList(nullable, source))
+        .processedWith(new AutoValueProcessor(Lists.newArrayList(new AutoValueGsonExtension())))
+        .compilesWithoutError()
+        .and()
+        .generatesSources(expected);
+  }
+
+  @Test
+  public void simpleWithAnnotatedBuilder() {
```
ZacSweers: If you wanted to just move this into the sample project as a code test there rather than this, I'd be in favor of starting that pattern with this PR. Seems way more maintainable

ZacSweers: Same with the other new tests below

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -834,93 +809,1294 @@ public void simple() {
...
+        + "      jsonWriter.name(realFieldNames.get(\"e\"));\n"
+        + "      if (object.e() == null) {\n"
+        + "        jsonWriter.nullValue();\n"
+        + "      } else {\n"
+        + "        TypeAdapter<String> string_adapter = this.string_adapter;\n"
+        + "        if (string_adapter == null) {\n"
+        + "          this.string_adapter = string_adapter = gson.getAdapter(String.class);\n"
+        + "        }\n"
+        + "        string_adapter.write(jsonWriter, object.e());\n"
+        + "
```
ZacSweers: Thanks for figuring this out and fixing the others!

example/src/test/java/com/ryanharter/auto/value/gson/example/PersonTest.java
```diff
@@ -39,4 +45,31 @@ public void testGson() throws Exception {
...
+    @Test
+    public void testGsonWithValidation() {
+        expectedException.expect(IllegalArgumentException.class);
+        expectedException.expectMessage("age cannot be negative");
+
+        Gson gson = new GsonBuilder()
+            .registerTypeAdapterFactory(SampleAdapterFactory.create())
+            .create();
+
+        String json = "{\"name\":\"Piasy\",\"gender\":1,\"age\":-1,\"birthdate\":\"2007-11-11\",\"address\":{\"street-name\":\"street\",\"city\":\"city\"}}";
```
ZacSweers: Fun fact - annotate this with `@Language("json")` and intelliJ will syntax highlight it. Or put `//language=json` on the line above

ZacSweers:
Also CI fix worked. Ignore the gcloud build, assuming that's one of @rharter's experiments :)

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: `BuilderContext#builderMethods` only returns static no-arg factory methods, so shouldn't be an issue. But perhaps worth calling out in the README?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: Good call, I found this annoying too.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: It does. I think I could also do the following. Not sure which is better practice.
```suggestion
      readMethod.addStatement("$T $N = $T.$N()", builder.type, builder, autoValueClassName, builderMethod.name);
```

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
bryanstern: Ya, let me take a stab at that

rharter:
> Also CI fix worked. Ignore the gcloud build, assuming that's one of @rharter's experiments :)

Hmm, not sure how that ended up on this repo.  Agreed, ignore that.

README.md
```diff
@@ -79,6 +79,32 @@ shared transience annotations library and is an `api` dependency of the runtime
...
+  public static Builder builder() {
+    return new AutoValue_Foo.Builder();
+  }
+
+  @AutoValueGsonBuilder
+  public static Builder builderWithDefaults() {
+    return new builder().quux("QUUX");
+  }
+  
+  public static Builder builder() {
```
rharter: duplicated method from above.

bryanstern: 🤦‍♂

rharter:
@zacsweers any thoughts?

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
ZacSweers: Gotcha, yeah a readme bit would be good

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
ZacSweers: We can leave it as-is :+1:

PR #227, opened 2019-10-16
Fix factory method with generics

### LINKED ISSUE #226: Build failed in the generated read method for a generic class
https://github.com/rharter/auto-value-gson/blob/a1aa36ebcd3baf770840fc3c603f6db328e7ca24/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java#L677
The line above will generate an error in the generated code for a generic class.
`error: <identifier> expected
      ObjectResponse.Builder builder = ObjectResponse<T>.<T>builder();`
```
    @Override
    @SuppressWarnings("unchecked")
    public ObjectResponse<T> read(JsonReader jsonReader) throws IOException {
      if (jsonReader.peek() == JsonToken.NULL) {
        jsonReader.nextNull();
        return null;
      }
      jsonReader.beginObject();
      ObjectResponse.Builder builder = ObjectResponse<T>.<T>builder(); // Build failed

      ...
```

version: 1.1.0

bryanstern:
Fixes #226.

Invalid java code was being generated when trying to instantiate a builder with generics via its factory method inside the `TypeAdapter#read`.

The fix is to use the builder's type rather than class when declaring the builder field inside of `TypeAdapter#read`.

I've also updated `GenericsExample` to show that it works and to serve as a test. The read method generated now looks like.

```java
public GenericsExample<A, B, C> read(JsonReader jsonReader) throws IOException {
  if (jsonReader.peek() == JsonToken.NULL) {
    jsonReader.nextNull();
    return null;
   }
   jsonReader.beginObject();
   GenericsExample.Builder<A, B, C> builder = GenericsExample.builder();
   ...
}
```

### DIFF
--- a/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -451,8 +451,8 @@ public class AutoValueGsonExtension extends AutoValueExtension {
         .addMethod(constructor.build())
         .addMethod(createWriteMethod(autoValueTypeName, properties, adapters,
             jsonAdapter, typeParams))
-        .addMethod(createReadMethod(className, autoValueTypeName, properties, adapters,
-            jsonAdapter, typeParams, builderContext, processingEnvironment));
+        .addMethod(createReadMethod(className, autoValueClassName, autoValueTypeName, properties,
+            adapters, jsonAdapter, typeParams, builderContext, processingEnvironment));
 
     if (!typeParams.isEmpty()) {
       classBuilder.addField(FieldSpec.builder(Type[].class, "typeArgs", PRIVATE, FINAL).build());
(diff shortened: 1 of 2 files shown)


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -589,7 +590,7 @@ private MethodSpec createReadMethod(ClassName className,
             .addMember("value", "\"unchecked\"")
             .build())
         .addModifiers(PUBLIC)
-        .returns(autoValueClassName)
+        .returns(autoValueTypeName)
```
bryanstern: This isn't functionally any different. Just a renamed variable.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -674,8 +675,8 @@ private MethodSpec createReadMethod(ClassName className,
           }
         }
 
-        readMethod.addStatement("$T $N = $T.$L", builderField.get().type, builderField.get(),
-            autoValueClassName, builderMethod);
+        readMethod.addStatement("$T $N = $T.$N()", builderField.get().type, builderField.get(),
+            autoValueClassName, builderMethod.getSimpleName());
```
bryanstern: `autoValueClassName` this is now a `ClassName` instance rather than a `TypeName` instance. We needed a `ClassName` instance otherwise this would have generated `Foo.Builder<T> builder = Foo<T>.builder();`

PR #228, opened 2019-10-19
Make generated class + constructor package private

### LINKED ISSUE #219: Generated TypeAdapterFactory is public
The generated `AutoValueGson_FooTypeAdapterFactory` is public, although it seems that it must be package-private and only be used by `FooTypeAdapterFactory#create`.

Factory base class code:
```java
@GsonTypeAdapterFactory
public abstract class FooTypeAdapterFactory implements TypeAdapterFactory {
  
  // If I understand correctly, the clients are suppossed to use this
  // public and documented method, and not have direct access to
  // the generated class. The readme confirms such expectation.
  public static TypeAdapterFactory create() {
    return new AutoValueGson_FooTypeAdapterFactory();
  }
}
```

AutoValueGson version: 1.0.0

ZacSweers:


ZacSweers:
The runtime change is backward compatible to old generated code, but new generated code requires the new runtime version. I feel like it's acceptable, but we should call this out in the changelog. Lmk your thoughts

rharter:
I don't think this is quite right.  It looks like the OP was referencing the [Factory](https://github.com/rharter/auto-value-gson#factory) generated by the `@GsonTypeAdapterFactory` annotation, not the adapter generated by the `@GenerateTypeAdapter` annotation (the different mechanisms are quite confusing).  It doesn't make sense for the adapter generated specifically to be accessed from the `com.ryanharter.autovalue.gson` package to be package private.

ZacSweers:
Oh snap, you're right 😬. That said, any reason not to make this match the
visibility of the model?

rharter:
I'm a bit curious why you'd want this, and the model, to be package private if you also want it accessible from the `GenerateTypeAdapter.FACTORY` class.  Seems like setting visibility to something that doesn't match the intended use case just so the factory can reflectively modify the accessibility to violate that visibility.

Unless there's a reason that you'd want a top level adapter that **doesn't** involve using that factory, like you have your own factory in that package, but that violate's AutoValue's value of hiding the implementation details (the `AutoValue_` generated classes) behind static factory methods so they're easily interchangeable.

Is there some use case I'm missing?

ZacSweers:
This was the intended design when I wrote it (the internal version I wrote at Uber worked the same way iirc), with the intention that reflection handles the package private access and allow the adapter itself to remain an implementation detail as far as other source code is concerned.

ZacSweers:
Yeah for somewhere like Moshi it made sense to keep them public or match visibility since there's no factory option, but in this case the thought was that if you're using the factory support then you're already not using the `@GenerateTypeAdapter` annotation. Does that sound clearer?

rharter:
Yeah, that works for me.

PR #230, opened 2019-10-19
Make field name policy support opt-in

### LINKED ISSUE #208: Make field naming strategy support opt-in
This comes at a cost in both generated code and also makes the generated code not proguardable due to reflective field lookups. The encouraged solution would be to just use `@SerializedName`

ZacSweers:
Resolves #208. 

Right now we always generate code that eagerly resolves and uses the input gson's field name policy. This is a lot of overhead for something that isn't always necessary, so this gates it behind a `autovaluemoshi.useFieldNamePolicy` processor option.


README.md
```diff
@@ -100,6 +100,20 @@ useful for setting default values.
 }
 ```
 
+## Field name policy
+
+If you want the generated adapter classes to use the input `Gson` instance's field name policy, you can 
+enable this via `autovaluemoshi.useFieldNamePolicy` processor option. This acts as a flag (any value is ignored)
```
rharter: `autovaluegson`

ZacSweers: d'oh. Fixed in de6eeff

PR #231, opened 2019-12-18
Generate human readable helper toString() methods

ZacSweers:
Useful for debugging to know what kind of adapter you're using. Adopted from Moshi with a similar StringBuilder approach to reduce string pool bloating

Example:

```java
@Override
public String toString() {
  return new StringBuilder().append("TypeAdapter(").append("GenerateTypeAdapterTest.Taco").append(")").toString();
}
```


ZacSweers:
Need to update tests, but will wait for review of the implementation first before doing so

ZacSweers:
Eh just went ahead and updated the tests

PR #234, opened 2019-12-20
Fix missing originating element in generated adapters

### LINKED ISSUE #233: Originating element not set
With e.g.

```java
@GenerateTypeAdapter
@AutoValue
public abstract class Foo {
    abstract String bar();
    @SerializedName("Baz") abstract String baz();
    abstract int quux();
    abstract String with_underscores();

    // The public static method returning a TypeAdapter<Foo> is what
    // tells auto-value-gson to create a TypeAdapter for Foo.
    public static TypeAdapter<Foo> typeAdapter(Gson gson) {
        return new Foo_GsonTypeAdapter(gson);
    }
}
```

and dependencies

```gradle
implementation group: 'com.google.auto.value', name: 'auto-value-annotations', version: '1.7'
annotationProcessor("com.ryanharter.auto.value:auto-value-gson:1.1.1")
implementation("com.ryanharter.auto.value:auto-value-gson-runtime:1.1.1")
```

incremental annotation processing fails because it does not pass the originating element when generating `com.example.myapplication.Foo_GsonTypeAdapter`.

Both Gradle Java compilation and KAPT fail to run incrementally.

The issue seems to be that `AutoValueGsonExtension.generateClass` does not set originating element for the `TypeSpec.Builder` it is using to write the class.

ZacSweers:
Fixes #233 

I think before this Just Worked because the generated adapter was a nested type of a class that did have its originating element set, but when it's top-level it does need it set anyway. This sets it for both just to be doubly sure.

CC @gavra0


auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -741,7 +743,7 @@ public void simpleNoFieldNamePolicy() {
         + ".getType()).getType()).getType()).getType()));\n"
         + "        }\n"
         + "        "
-        + "map__string_map__string_map__string_map__string_map__string_wildcard__string_adapter"
```
rharter: nit: looks like the formatting got messed up in this file.

ZacSweers: I'm not sure really what to do about that, I just regenerated this one and intelliJ wants to line chomp it now :|

gavra0:
In 1.3.70 KAPT will have better logging regarding this. To figure out what is causing it to fail, you can create a breakpoint in https://github.com/JetBrains/kotlin/blob/a9686c94474c9d442db61b45f60e6ff326fdb7db/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/incrementalProcessors.kt#L119 (this is tag v1.3.61).

ZacSweers:
I've figured it out. This is a bug in JavaPoet, where calling `TypeSpec#toBuilder()` leaves out the originating elements.

https://github.com/square/javapoet/issues/749

auto-value-gson/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
bryanstern: I agree, preserve the indenting in spite of intelliJ.

PR #238, opened 2020-01-16
Extract factory to its own artifact

ZacSweers:
This helps incremental annotation processing by making the factory processor an optional dependency (and thus not strictly present on the processing classpath). Not strictly a breaking change but does require adding the explicit dependency now.


auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -176,12 +176,21 @@ public boolean applicable(Context context) {
...
         .getOptions()
         .containsKey(USE_FIELD_NAME_POLICY);
+    return isApplicable(context.autoValueClass(), context.processingEnvironment().getMessager());
+  }
+
+  public static boolean isApplicable(TypeElement type, Messager messager) {
+    @SuppressWarnings("UnstableApiUsage")
+    boolean generateExternalAdapter =
+        MoreElements.isAnnotationPresent(type, GenerateTypeAdapter.class);
+    if (generateExternalAdapter) {
```
bryanstern: Inline?

ZacSweers:
Not really, it's more that the shear presence of an aggregating processor on the classpath during incremental annotation processing is enough to trigger it to run in an aggregating mode, even if no annotations it looks for are used.

auto-value-gson/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
ZacSweers: 5895da2

PR #239, opened 2020-01-17
Introduce legacy artifact for existing users

ZacSweers:
This separates out the existing `auto-value-gson` artifact to be `auto-value-gson-extension` and just makes a shim `auto-value-gson` artifact for legacy reasons that just delegates to both.

Opportunistically enabled github actions like we have on auto-value-moshi. Commented out the sonatype uploads for now but left for future use


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Gradle Wrapper Validation
+        uses: gradle/wrapper-validation-action@v1
+      - name: Install JDK ${{ matrix.java_version }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
+      - name: Build with Gradle
+        run: ./gradlew clean check --stacktrace
+#      - name: Upload snapshot (master only)
+#        run: ./gradlew uploadArchives -PSONATYPE_NEXUS_USERNAME=${{ secrets.SonatypeUsername }} -PSONATYPE_NEXUS_PASSWORD=${{ secrets.SonatypePassword }}
```
rharter: I've added the secrets so you should be able to uncomment these.

README.md
```diff
@@ -208,14 +208,18 @@ new GsonBuilder()
...
 ```kotlin
-annotationProcessor("com.ryanharter.auto.value:auto-value-gson:1.2.0")
+annotationProcessor("com.ryanharter.auto.value:auto-value-gson-extension:1.2.0")
 implementation("com.ryanharter.auto.value:auto-value-gson-runtime:1.2.0")
 
 // Optional @GsonTypeAdapterFactory support
 annotationProcessor("com.ryanharter.auto.value:auto-value-gson-factory:1.2.0")
+
+// Legacy generic artifact that includes both -extension and -factory above. Should not be used
+// unless necessary.
```
rharter: Do you want to add a note explaining that you give up incremental compilation if you use this?  The "don't use this" note sounds ominous.

ZacSweers: Good point, will add something. Incremental compilation works, it's just aggregating instead of isolating

rharter: Oh, I thought that incremental depended on isolating processors, so this broke it.  That's good that it doesn't break incremental compilation, though.

.github/workflows/ci.yml
ZacSweers: Nice! 2bcf03e

README.md
ZacSweers: 2bcf03e

PR #243, opened 2020-01-20
Update gradle & shadow plugin to latest

ZacSweers:


ZacSweers:
Filed an issue for the shadow plugin exception here: https://github.com/johnrengelman/shadow/issues/540

PR #246, opened 2020-01-30
Fix generated proguard names for inner classes

ZacSweers:
Proguard speaks reflection names


ZacSweers:
The avg names were suspiciously passing, but turns out it's because compile-testing doesn't actually compare file paths. Will file there

image

PR #259, opened 2021-05-26
Generalize support for setting properties with builders.

eamonnmcmanus:
AutoValue has a number of ways to set a property with a method whose parameter
does not exactly match the return type of the property method. For example, a
property defined by `abstract ImmutableList<String> getFoo()` can be set by an
abstract method `abstract Builder setFoo(List<String>)`. AutoValue knows how
to implement this by calling `ImmutableList.copyOf`. Here, we don't need to know
about `ImmutableList.copyOf`, we just need to know that this setter is indeed
callable to set the value of the list. There are similar considerations for
list builders (`ImmutableList.Builder<String> fooBuilder()`) and for `Optional`
(`Optional<String>` set by plain `String`). This change allows auto-value-gson
to use these alternative ways of setting a property, rather than requiring that
the setter parameter exactly match the property type.

Also made two changes that should facilitate updating to newer JDKs and
to newer versions of the Truth library:

* use an appropriate `@Generated` annotation depending on the target
  Java version;

* remove a reference to the `named(String)` method, which no longer
  exists in Truth.

### DIFF
--- a/auto-value-gson-extension/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
+++ b/auto-value-gson-extension/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
@@ -48,7 +48,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.annotation.processing.Filer;
 import javax.annotation.processing.Messager;
@@ -58,9 +57,10 @@ import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.TypeParameterElement;
+import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
-import javax.lang.model.util.ElementFilter;
+import javax.lang.model.util.Types;
 import javax.tools.Diagnostic;
 
 import static com.google.common.base.CaseFormat.LOWER_CAMEL;
@@ -72,6 +72,7 @@ import static javax.lang.model.element.Modifier.PRIVATE;
 import static javax.lang.model.element.Modifier.PUBLIC;
 import static javax.lang.model.element.Modifier.STATIC;
 import static javax.lang.model.element.Modifier.VOLATILE;
+import static javax.lang.model.util.ElementFilter.methodsIn;
 
 @SupportedOptions(USE_FIELD_NAME_POLICY)
 @AutoService(AutoValueExtension.class)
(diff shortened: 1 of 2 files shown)


auto-value-gson-extension/src/test/java/com/ryanharter/auto/value/gson/AutoValueGsonExtensionTest.java
```diff
@@ -34,6 +36,10 @@
 
 @RunWith(JUnit4.class)
 public class AutoValueGsonExtensionTest {
+  private static final String GENERATED =
```
ZacSweers, line 39: Nice

auto-value-gson-extension/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
```diff
@@ -555,24 +556,122 @@ private static void addBuilderFieldSetting(Builder block,
...
+      }
+    }
+    // This shouldn't happen: AutoValue should have rejected the builder if it didn't have a way to
+    // set this property. But it may be that AutoValue has output an error message and still let
+    // this extension run.
+    String errorMsg = "Setter not found for " + prop.element;
+    processingEnv.getMessager().printMessage(Diagnostic.Kind.ERROR, errorMsg, prop.element);
+  }
+
+  private static Optional<TypeMirror> optionalArgType(TypeMirror typeMirror, Types typeUtils) {
```
ZacSweers, line 653: It's worth mentioning we don't technically support `Optional` properties in auto-value-gson (#161, #123, https://github.com/google/auto/issues/359). If it's working, it's by coincidence and not intentional :). Does that obviate the need for much of the below or do you think it's worth keeping this around as a toe-hold for better supporting it in the future?

ZacSweers (approved):
Looks great! Thanks for doing this and apologies for the delay in review. Dropped one question that I'd be curious for thoughts on before we merge.

Also - this repo hasn't been released for awhile as it's in sort of maintenance mode. Does having this merged onto `main` work for y'all's internal needs or would you want to see a release after this?

auto-value-gson-extension/src/main/java/com/ryanharter/auto/value/gson/AutoValueGsonExtension.java
eamonnmcmanus: The background to this PR is that I am trying to update Google's giant internal [monorepo](https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext) to a more recent version of auto-value-gson. We're currently on v0.8.0 so a lot has changed since then. I've basically imported the current code to my own workspace, and I'm working on fixing the things that that change breaks before I can submit it to the monorepo. The version we have predates the use of builders during deserialization, so some things that happened to work then don't work now. This PR is an attempt to address some of those things, in a way that is useful in general and not just to Google.

I haven't really studied the question of `Optional` specifically in detail, but this change does help with some of the cases I just mentioned. Yes, google/auto#359 would make some of the code here a lot simpler if it were addressed, but there's some API review and release choreography that would have to happen before we could make that simplification.

In answer to your other question, I'm importing the code from the current `main` branch. I don't need a release for this to be useful to me.

ZacSweers: Great, I'll merge this then. @cushon lookee into this awhile back iirc and ran into issues with our past removal of mutable adapters + default value setting, fyi

eamonnmcmanus: Thanks! Yes, I'm running into those same issues too, but I think I'll be able to fix them by modifying the application code in question rather than by tweaking auto-value-gson.

PR #262, opened 2021-06-12
Fix `TYPE_USE` annotations on nested types.

### LINKED ISSUE #261: `@Nullable` doesn't work correctly with nested types
If the `@AutoValue` class has a property like this:
 
```
  abstract Map.@Nullable Entry<?,?> bar();
```

then auto-value-gson incorrectly generates a constructor like this:

```
  Foo(@Nullable Map.Entry<?,?> bar) {...}
```

It needs to use the same `Map.@Nullable Entry` syntax.

I am preparing a PR to fix this.

eamonnmcmanus:
Previously if you had a property of type `Map.@Nullable Entry`
then it would end up in the generated source code as
`@Nullable Map.Entry`, which doesn't compile.

Fixes #261.


ZacSweers (approved):
Thanks!

joaolbarba88:
>

joaolbarba88:
Desbloquear algoritmo Uber

